PR #83, opened 2020-10-15
Support for Prysm voluntary exit

yorickdowne:


yorickdowne:
Please do NOT approve this yet, let's discuss. Master just got a new flag: `--accept-terms-of-use`. I am thinking to put that into `.env` with a link so the user has to manually choose "true" with an edit. Just accepting on behalf of the user feels wrong.

Feedback?

tjcim (approved):
Looks good.

PR #86, opened 2020-10-16
Changed Prysm syntax: accounts-v2 is now accounts

yorickdowne:


yorickdowne:
Skipping review. The change is small, tested, and necessary because the current release breaks with Prysm `master`

PR #95, opened 2020-11-05
Fix Prysm Grafana

yorickdowne:


danbryan (approved):
Looks good to me, nice one.

PR #111, opened 2020-11-13
adding nethermind as eth1

adrienlacombe:


yorickdowne (changes requested):
dotnet is not a prerequisite on the host, docker takes care of that. Please revert this change

yorickdowne (changes requested):
For README.md, looks good, but "eth1" instead of "eht1" please.

yorickdowne (changes requested):
default.env changes look mostly good, please revert the commenting out of GETH1_NETWORK. It doesn't hurt you when using Nethermind as you'd most likely be referencing ETH1_NETWORK instead.

yorickdowne (changes requested):
Dockerfile needs a bit of work. You're close.

The BUILD_TARGET gets referenced in clone, this is where you decide which branch to grab.

ETH1_NETWORK needs to be something that can be specified at runtime, not just compile time.

The part at "this is no longer from NM github" is usually the final stage, and copies executables from builder.

nm.yml looks mostly good, I think you want the jsonrpc port to be ETH1_RPC_PORT, and as discussed above, ETH1_NETWORK should be able to be specified during runtime.

PR #119, opened 2020-11-16
Binary client option

yorickdowne:
Putting this here for review instead of waiting until all clients have received that option


danbryan (approved):
Discussed in discord, otherwise looks good to me!

PR #160, opened 2020-12-06
Improved documentation

yorickdowne:


danbryan (approved):
Looks good, firewall is well documented, and the cloud vps is explained well, and i think its great you put in some info about saving the mnemonic.

PR #166, opened 2020-12-22
Allow access to critical ports only from localhost

pico-prob:
Docker and ufw both use `iptables` to manage networks. IMHO Docker rules overrride the ufw rules that secure the host. Sometimes it helps just to `sudo ufw reload` after the container start to resolve the issue. But this is not always the case.

There's a [long discussion](https://github.com/docker/for-linux/issues/690) on how to deal with this highly critical issue. Surprisingly for years without a canonical approach.

A very simple/straight forward way to solve this is to **restrict the docker network to expose specific ports only to localhost**:

Example:
<pre>
    ports:
      - ${GRAFANA_PORT}:3000/tcp
</pre>
... changes to: 
<pre>
    ports:
      - <b>127.0.0.1:</b>${GRAFANA_PORT}:3000/tcp
</pre>

I did this for all ports that are noted as 'sensitive' in the `default.env`. Tested successfully for `prysm-grafana.yml`

### DIFF
--- a/eth1-shared.yml
+++ b/eth1-shared.yml
@@ -1,7 +1,7 @@
 # To be used in conjunction with oe.yml, nm.yml, besu.yml or geth.yml
-version: "3.4"
+version: '3.4'
 services:
   eth1:
     ports:
-      - ${ETH1_RPC_PORT}:${ETH1_RPC_PORT}/tcp
-      - ${ETH1_WS_PORT}:${ETH1_WS_PORT}/tcp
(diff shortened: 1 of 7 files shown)


yorickdowne:
I like it, this is potentially much cleaner than forcing ufw into the chain. It'd break with eth1 standalone however, I think. More testing needed before this is accepted. It'd also restrict grafana in non-cloud environments, though we're about to solve that another way with traefik.

yorickdowne:
Closing this in favor of using traefik. People who do not want traefik can still use ufw in the chain instead.

PR #168, opened 2020-12-26
docs(README): use updated Prysm dashboard JSON

blankey1337:
The dashboard JSON was updated recently to support "beacon node" instead of "beacon". This is the correct dashboard URL to use with eth2-docker.


mfiumara:
Thanks for this PR this fixes some issues where some of the graphs in the dashboard are not populated.

PR #180, opened 2021-01-13
updated to latest eth2-docker

danbryan:
updated to latest eth2-docker


danbryan:
pulled wrong way, deleting

PR #207, opened 2021-03-09
Changes to deposit-cli UX

yorickdowne:
In docs, we'd use `sudo docker-compose run --rm deposit-cli-add-recover --folder seed_check`  for verifying the seed phrase.

`deposit.yml` exists in this but hasn't replaced what's in the client-base.yml files. That could be a separate PR if we think that UX is better.


danbryan (approved):
Looks good!

PR #270, opened 2021-07-20
Enable Prysm doppelganger protection by default

yorickdowne:


mfiumara:
Hey from the release notes of Prysm this is indicated still as experimental and will become default in a later release, why include this as default now?

yorickdowne:
Only because there was one user, not on eth-docker, who got themselves slashed, which would have been prevented by this option. I am erring on the side of caution.

PR #289, opened 2021-09-06
Added note about Altair fork

currentoor:
The latest stable version of Prysm does not work with the Prater testnet. Need to use the latest unstable release.

https://github.com/prysmaticlabs/prysm/releases/tag/v2.0.0-rc.1

But I'm not sure if this should go in the docs or here.

### DIFF
--- a/default.env
+++ b/default.env
@@ -108,6 +108,11 @@ LH_DOCKER_TAG=latest
 LH_DOCKERFILE=Dockerfile.binary
 
 # Prysm
+# If running on Prater testnet you need to use v2.0.0-rc.1 because of the Altair fork.
+# https://github.com/prysmaticlabs/prysm/releases/tag/v2.0.0-rc.1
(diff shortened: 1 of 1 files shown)


yorickdowne:
I'm disinclined to put that into docs, since it's such a short-term situation. That said, having a way to signal that would be good. Maybe changelog?

currentoor:
I don't see a CHANGELOG.md file. Perhaps you meant something else. I'm fine with whatever you think is best.

I could open an Issue instead if you'd like, you can close it when this patch is no longer necessary.

yorickdowne:
Changelog is here: https://eth-docker.net/docs/About/Changelog

Prysm is expected to release an Altair-ready version this week, alongside the other three clients.

currentoor:
Ah, I see, in that case, this doesn't seem necessary.

PR #299, opened 2021-09-29
Support new beaconcha.in stats endpoint - separate consensus and validator

fmoledina:
Carbon copy of PR #273 for the separate lh-stats-consensus.yml and lh-stats-validator.yml components of the stack.

### DIFF
--- a/lh-stats-consensus.yml
+++ b/lh-stats-consensus.yml
@@ -4,4 +4,4 @@ services:
   consensus:
     command:
       - --monitoring-endpoint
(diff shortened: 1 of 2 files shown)


fmoledina:
Also included an updated note in default.env regarding applicability of beaconcha.in stats env vars to both Prysm and Lighthouse.

fmoledina:
Updated PR to remove changes to default.env as the have been captured upstream.

yorickdowne:
Thank you!

PR #471, opened 2022-02-09
Add promtail/loki for getting all container logs to grafana

gathecageorge:
After starting all services, one can view logs by tailing the specific containers or compose as from the documentation at 

https://eth-docker.net/docs/Usage/Deposit#:~:text=docker%2Dcompose%20logs%20%2Df%20consensus

This feature allows viewing of logs directly from grafana if enabled. You can choose which container to view logs for and even search and filter using Explore menu in grafana

### DIFF
--- a/besu.yml
+++ b/besu.yml
@@ -5,6 +5,7 @@ x-logging: &logging
     options:
       max-size: 10m
       max-file: "3"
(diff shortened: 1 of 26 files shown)


yorickdowne:
This fails lint. Please add that locally and run a commit to let the linter fix things. Looks like end of file complains. See https://github.com/eth-educators/eth-docker/blob/main/CONTRIBUTING.md for instructions to add the lint to pre-commit.

PR #756, opened 2022-07-20
Support git 1.8.3

joeytwiddle:
Although `git branch --show-current` is certainly cleaner, it fails on some OSes which haven't updated to git 2.

For wider compatibility, `git rev-parse --abbrev-ref HEAD` produces the same result. ([source](https://www.techiedelight.com/determine-current-branch-name-git/))

It's just ugly for developers!

### DIFF
--- a/ethd
+++ b/ethd
@@ -82,7 +82,7 @@ update() {
         esac
     done
 
-    if ${__as_owner} git branch --show-current | grep -q rpc-nodes; then
+    if ${__as_owner} git rev-parse –abbrev-ref HEAD | grep -q rpc-nodes; then
         __rpc_branch=1
     else
         __rpc_branch=0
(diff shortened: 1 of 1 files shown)


yorickdowne:
Interesting. What's the OS you're on, there?

yorickdowne:
Pondering https://stackoverflow.com/questions/47098342/jenkinsfile-git-rev-parse-abbrev-ref-head-returns-head . If someone is on a commit ("detached head") we'd get the wrong answer. Bit niche to be sure ..

yorickdowne:
I don't have a git 1.8.3. This could work: `git name-rev --name-only HEAD`

Can you test that when you are on the branch and when you're in detached head state on a previous commit?

yorickdowne:
Change made.

PR #758, opened 2022-07-20
Check for whiptail, and working docker, before starting

joeytwiddle:
Current behaviour:

- When `whiptail` is not present, `ethd config` exits with "You selected Cancel"

I think it should exit with a clear error message instead.

First I tried adding `set -e` but that just caused the script to exit with no output at all! Which doesn't really help the user.

So instead, I added a little check for `whiptail` when the script starts up.

(On CentOS 7.9, there is no `whiptail` package, but it can be found in the `newt` package.)

I also added a check that the user can call docker.

I can refactor to use `$?` if you need it. I just use `if` directly.

### DIFF
--- a/ethd
+++ b/ethd
@@ -1183,6 +1183,11 @@ if [ "${OWNER}" == "root" ]; then
     exit 1
 fi
 
+if ! type -P whiptail >/dev/null 2>&1; then
+    echo "Please install the package whiptail or newt before running this script"
+    exit 1
+fi
+
 ENV_FILE=.env
 
 command="$1"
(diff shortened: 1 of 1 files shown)


ethd
```diff
@@ -1192,4 +1197,9 @@ determine_distro
 determine_compose
 handle_root
 
+if ! docker images >/dev/null 2>&1; then
+    echo "Please ensure your user can call docker before running this script"
+    exit 1
+fi
+
```
joeytwiddle, line 1204: Originally I had wanted to test:

```
if ! cmd ls >/dev/null 2>&1
```

But `docker-compose ls` doesn't work until the config has been generated, so this would always fail on the first run!

I couldn't find a `docker-compose` command that would run without config, but fail if docker could not be accessed.

So I just test `docker images` instead.

yorickdowne:
Good ideas, thank you!

PR #760, opened 2022-07-20
Automatically use sudo if needed

joeytwiddle:
If I'm not mistaken, this would be needed if the user does not have access to call docker.

(And I think using sudo is probably preferable to giving the user access to docker, because as the docs say, this gives root access without a password.)

### DIFF
--- a/ethd
+++ b/ethd
@@ -37,6 +37,11 @@ determine_compose() {
   else
     __compose_exe="docker-compose"
   fi
+
+  # If this user cannot access docker directly, then use sudo
+  if ! docker images >/dev/null 2>&1; then
+    __compose_exe="sudo $__compose_exe"
+  fi
 }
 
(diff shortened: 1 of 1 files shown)


yorickdowne:
Hmm. Feels almost like taking agency from the user, springing sudo on them ... then again it'd simplify the documentation, and they can still decide whether they want to be part of `docker` group, or not.

joeytwiddle:
I see what you mean about taking agency away from the user.

An alternative solution might be to make it explicit:

1. If the user passes `--use-sudo` or `-s` then we add `sudo` to the `__compose_exe`

2. if they do not do that, and docker cannot be accessed by the user, then we suggest the user to do that, and exit.

If you prefer this solution, I can write it!

yorickdowne:
I like that idea. Please write it!

PR #781, opened 2022-07-26
Use v2 GetState API for Teku checkpoint sync

jimmygchen:
v1 phase0 endpoint is deprecated and v2 should be used post Altair.

When I started Teku, the following URL returns 404
https://goerli.checkpoint-sync.ethdevops.io/eth/v1/debug/beacon/states/finalized

Changing to v2 fixed this.


yorickdowne:
Thanks!

PR #784, opened 2022-07-27
Only escalate with sudo if user requests it

joeytwiddle:
Improvement for #760

### DIFF
--- a/ethd
+++ b/ethd
@@ -38,10 +38,7 @@ determine_compose() {
     __compose_exe="docker-compose"
   fi
 
-  # If this user cannot access docker directly, then use sudo
-  if ! docker images >/dev/null 2>&1; then
-    __compose_exe="sudo $__compose_exe"
-  fi
+  __compose_exe="$__maybe_sudo $__compose_exe"
 }
 
 handle_root() {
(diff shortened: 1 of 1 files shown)


joeytwiddle:
I find it a little inconvenient for the user to type `ethd -s` each time. 

A friendlier way might be to ask/check during the config stage, store the decision in the `.env` file, and then fetch the setting from that file on future calls.

I could try that approach, as long as you don't mind me doing `source .env` hehe!

yorickdowne:
Not a fan of `source .env`, it goes horribly wrong. What I've been doing is iterate through it with `sed` to get values. 

Could be a thing where config detects whether sudo is needed: Yes if not on macOS and user ... let's say owner here because ethd itself could be run sudo during config ... is not part of docker group. A variant of the check at the bottom of config(), just reversed. And in that case ask the user whether they'd like auto-sudo.

The check at the bottom of config() would then also need to check whether to use auto-sudo, since that's another case where running the build after config is appropriate.

yorickdowne:
On further thought. I think escalating to sudo automatically is fine. Running docker commands does indicate the user wants to, well, run those. I am closing this. Thanks for the contributions!

PR #796, opened 2022-07-30
Add docker-compose version check

aliask:
Updating my Goerli node, I found that things stopped working, and I got unexpected errors about unsupported configuration items in `prysm-base.yml`.

I did some investigation and it appears that the `profiles` option which was introduced in 0e2447abd5031975a2e20c43b919f2aba4c36ce5 means that at least `docker-compose` version 1.28.0 is needed and mine was too old - Ubuntu version from default apt is 1.25.0.

This PR adds a simple version check to `determine_compose()`, similar to the Ubuntu version test.

### DIFF
--- a/ethd
+++ b/ethd
@@ -35,6 +35,13 @@ determine_compose() {
   if [ $? -ne 0 ]; then
     __compose_exe="docker compose"
   else
+    __compose_version=$(docker-compose --version | grep -oP "version \K[\w.-]+")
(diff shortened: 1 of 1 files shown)


yorickdowne:
I like it. It'll fail on macOS though. Could you take a look at the `sed` replacement for `grep -oP` in ethd update, and do something similar here? With that change it'll be ready to merge.

PR #835, opened 2022-08-06
Fix numeric range for docker-compose version check

aliask:
Docker-compose versions can include 0, so we need to make sure they're included in case a `1.30` is released.

```shell
aliask@ubuntu:~/eth-docker$ cat test.sh
#!/bin/bash
__compose_version=$(echo "docker-compose version 1.30.0" | sed -n -e "s/.*version \([1-9.-]*\).*/\1/p")
__compose_version_major=$(echo $__compose_version | cut -f1 -d.)
__compose_version_minor=$(echo $__compose_version | cut -f2 -d.)

echo "$__compose_version_major.$__compose_version_minor"
aliask@ubuntu:~/eth-docker$ ./test.sh
1.3
```

### DIFF
--- a/ethd
+++ b/ethd
@@ -35,7 +35,7 @@ determine_compose() {
   if [ $? -ne 0 ]; then
     __compose_exe="docker compose"
   else
-    __compose_version=$(docker-compose --version | sed -n -e "s/.*version \([1-9.-]*\).*/\1/p")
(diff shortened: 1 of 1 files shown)


yorickdowne:
Merging this, and. there won't be any further 1.x. Afaik everything is now `docker compose`, the plugin, version 2.x

PR #861, opened 2022-08-10
Set the host IP to use for containers

romanoLT:
Some dedicated server have a full range of IP allocated to them. This pull request gives the option for the user to specify which IP address from the Host will be set as default for eth-docker.
In Docker, by default when you run a container, all the host IPs point to the container. With this proposal, the user is able to specify the IP to use so only one IP redirects from the outside to the Docker container.
Tested for IPv4.


yorickdowne:
This should default to `0.0.0.0` when not set, as in `${HOST_IP:-0.0.0.0}:${EL_P2P_PORT:-30303}:${EL_P2P_PORT:-30303}/tcp`

I am careful to only introduce what is needed. Typically access is controlled by firewalls. What is the use case here to restrict the docker mapping to just one IP?

romanoLT:
Hey @yorickdowne ,
It might not be the most common scenario but if you want to handle several Docker containers on your host using different IP you have to handle the IP allocation within Docker and not on the firewall as it would be to complex (many conditional rules etc...)
I'll reflect your comments.

yorickdowne:
ethd has changed since you prepped this. You're better off starting off with a fresh version of it and just adding `HOST_IP` to the `ALL_VARS`.

A couple files were renamed since. Shouldn't be too hard to clear up these conflicts. Close!

romanoLT:
making a new pull request with a merge-ready base commits

PR #863, opened 2022-08-10
Minor improvements

joeytwiddle:
A few small changes, hopefully for the better

### DIFF
--- a/default.env
+++ b/default.env
@@ -75,10 +75,10 @@ TRAEFIK_WEB_PORT=443
 TRAEFIK_WEB_HTTP_PORT=80
 # NOTE: If you configure EL_RPC_PORT or EL_WS_PORT then you should also add el-shared.yml to COMPOSE_FILE
 # Execution client RPC port, important when using ec-shared.yml. Insecure, do not expose to Internet.
-#EL_RPC_PORT=8545
+EL_RPC_PORT=8545
 # Execution client WS port. Insecure, do not expose to Internet.
 # Note that for Erigon, this needs to match EL_RPC_PORT
-#EL_WS_PORT=8546
+EL_WS_PORT=8546
 # Erigon's torrent port. Don't make this 42070, that'll fail
 ERIGON_TORRENT_PORT=42069
 # SSV Node ports
--- a/default.env
+++ b/default.env
@@ -73,11 +73,12 @@ KEY_API_PORT=7500
 # Secure web proxy port, 443 and 80 are great defaults
 TRAEFIK_WEB_PORT=443
 TRAEFIK_WEB_HTTP_PORT=80
+# NOTE: If you configure EL_RPC_PORT or EL_WS_PORT then you should also add el-shared.yml to COMPOSE_FILE
 # Execution client RPC port, important when using ec-shared.yml. Insecure, do not expose to Internet.
-EL_RPC_PORT=8545
+#EL_RPC_PORT=8545
 # Execution client WS port. Insecure, do not expose to Internet.
 # Note that for Erigon, this needs to match EL_RPC_PORT
-EL_WS_PORT=8546
+#EL_WS_PORT=8546
 # Erigon's torrent port. Don't make this 42070, that'll fail
 ERIGON_TORRENT_PORT=42069
 # SSV Node ports
--- a/ethd
+++ b/ethd
@@ -79,7 +79,7 @@ upgrade_compose() {
     if [ "$__compose_version_major" -eq 1 -a "$__compose_version_minor" -lt 28 ]; then
       echo "Found docker-compose version $__compose_version_major.$__compose_version_minor, upgrading to 1.29.2"
       ${__auto_sudo} apt-get install -y curl
-      ${__auto_sudo} curl -L "https://github.com/docker/compose/releases/download/1.29.2/docker-compose-$(uname -s)-$(uname -m)" -o /usr/bin/docker-compose
+      ${__auto_sudo} curl -fsSL "https://github.com/docker/compose/releases/download/1.29.2/docker-compose-$(uname -s)-$(uname -m)" -o /usr/bin/docker-compose
       ${__auto_sudo} chmod +x /usr/bin/docker-compose
     fi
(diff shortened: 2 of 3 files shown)


default.env
```diff
@@ -73,11 +73,12 @@ KEY_API_PORT=7500
...
 TRAEFIK_WEB_PORT=443
 TRAEFIK_WEB_HTTP_PORT=80
+# NOTE: If you configure EL_RPC_PORT or EL_WS_PORT then you should also add el-shared.yml to COMPOSE_FILE
 # Execution client RPC port, important when using ec-shared.yml. Insecure, do not expose to Internet.
-EL_RPC_PORT=8545
+#EL_RPC_PORT=8545
 # Execution client WS port. Insecure, do not expose to Internet.
 # Note that for Erigon, this needs to match EL_RPC_PORT
-EL_WS_PORT=8546
+#EL_WS_PORT=8546
```
joeytwiddle, line 81: I commented these out because they don't actually do anything until `el-shared.yml` is added.

Hopefuly the process of uncommenting them, will also encourage the user to read the note!

ethd
```diff
@@ -1351,14 +1351,14 @@ printhelp() {
     echo "Be sure to run the config command as a non-root user, and all other commands"
     echo "either with sudo, or as a user who is part of the \"docker\" group"
     echo ""
-    exit 0
 }
 
 if [[ "$#" -eq 0 || "$1" == "help" || "$1" == "-h" || "$1" == "--help" ]]; then
     printhelp
+    exit 0
```
joeytwiddle, line 1358: I think this is clearer than renaming `printhelp` to `print_help_and_exit` ;-)

ethd
```diff
@@ -1351,14 +1351,14 @@ printhelp() {
...
 fi
 
-cd $(dirname $(realpath ${BASH_SOURCE}))
+cd "$(dirname "$(realpath ${BASH_SOURCE})")"
```
joeytwiddle, line 1361: I think this might be useful for macOS users, where folders containing spaces are quite common.

E.g. `~/Library/Application Support/`

prysm/docker-entrypoint.sh
```diff
@@ -41,21 +41,21 @@ if [[ "$1" =~ ^(beacon-chain)$ ]]; then
...
-      curl -L -o "$GENESIS" https://github.com/eth-clients/merge-testnets/raw/main/ropsten-beacon-chain/genesis.ssz
+      curl -fsSL -L -o "$GENESIS" https://github.com/eth-clients/merge-testnets/raw/main/ropsten-beacon-chain/genesis.ssz
     fi
     exec "$@" "--genesis-state=$GENESIS" ${__rapid_sync} ${__override_ttd} ${__mev_boost}
   elif [[ "$@" =~ --sepolia ]]; then
     GENESIS=/var/lib/prysm/genesis.ssz
     if [ ! -f "$GENESIS" ]; then
       echo "Fetching genesis file for Sepolia testnet"
-      curl -L -o "$GENESIS" https://github.com/eth-clients/merge-testnets/raw/main/sepolia/genesis.ssz
+      curl -fsSL -L -o "$GENESIS" https://github.com/eth-clients/merge-testnets/raw/main/sepolia/genesis.ssz
```
joeytwiddle, line 58: There are more calls to `curl` which could use `-fsSL` but these are inside docker build scripts. I have left them alone for now.

joeytwiddle: Originally I used `--no-progress-meter` but I think `-fsSL` is a good default to use.

prysm/docker-entrypoint.sh
```diff
@@ -41,21 +41,21 @@ if [[ "$1" =~ ^(beacon-chain)$ ]]; then
...
-      curl -L -o "$GENESIS" https://github.com/eth-clients/merge-testnets/raw/main/ropsten-beacon-chain/genesis.ssz
+      curl -fsSL -o "$GENESIS" https://github.com/eth-clients/merge-testnets/raw/main/ropsten-beacon-chain/genesis.ssz
     fi
     exec "$@" "--genesis-state=$GENESIS" ${__rapid_sync} ${__override_ttd} ${__mev_boost}
   elif [[ "$@" =~ --sepolia ]]; then
     GENESIS=/var/lib/prysm/genesis.ssz
     if [ ! -f "$GENESIS" ]; then
       echo "Fetching genesis file for Sepolia testnet"
-      curl -L -o "$GENESIS" https://github.com/eth-clients/merge-testnets/raw/main/sepolia/genesis.ssz
+      curl -fsSL -o "$GENESIS" https://github.com/eth-clients/merge-testnets/raw/main/sepolia/genesis.ssz
```
joeytwiddle, line 58: Originally I added `--no-progress-meter` but I saw you had used `-fsSL` and I think these are good base options.

There are more calls to `curl` which could use `-fsSL` but these are inside docker build scripts. I left them alone for now.

default.env
yorickdowne: Not on board :)

ethd
yorickdowne: Sure

ethd
yorickdowne: Yeah that's useful

prysm/docker-entrypoint.sh
yorickdowne: Can't hurt.

default.env
yorickdowne: While I *think* I have defaults for these in all .yml files, I'd rather not rely on that. Let's not comment them out. The flow overall is that things used in yml files are defined in .env. Yes, I have added defaults to some stuff just in case, but I don't want to rely on it.

joeytwiddle: Understood. I have uncommented them again.

ethd
joeytwiddle: Honestly, I just put quotes around everything in bash. It's easier to do that, than remember when I do/don't need to!

I like to use [`shellcheck`](https://github.com/koalaman/shellcheck#user-content-in-your-editor) to help me write clean shell code. But it's easier to use that with new scripts. It always screams at my old scripts!

PR #864, opened 2022-08-10
[WIP] Sudo improvements

joeytwiddle:
WIP because I will test this either this week or next. I am posting it now, for feedback and review.

This cleans up some leftover issues with #760:

- PR 760 handled the calls to `docker-compose` but _not_ the calls to `docker`

----

We now have these two functions:

```bash
cmd() {
    $__maybe_sudo $__compose_exe "$@"
}

dodocker() {
    $__maybe_sudo docker "$@"
}
```

Thoughts:

1. We should probably name these functions consistently.

   One option is `docompose()` and `dodocker()` but you may have another preference.

2. The functions were _supposed_ to encapsulate the use of `$__maybe_sudo`

   But in fact there are some places in the script where we still need to use `$__maybe_sudo $__compose_exe` and `$__maybe_sudo docker` which looks a bit ugly.

   I could make those a bit cleaner, if I put the `sudo` back inside `$__compose_exe` and inside `$__docker_exe` variables.

   I think I will try that arrangement later. We can always rollback if it doesn't feel better.

   **Update:** I did that. It does look a bit cleaner now. But there arestill places where we need to write `$__maybe_sudo docker-compose` (before we have determined `$__compose_exe`).

### DIFF
--- a/ethd
+++ b/ethd
@@ -1399,12 +1399,12 @@ if ! type -P whiptail >/dev/null 2>&1; then
 fi
 
 if ! dodocker images >/dev/null 2>&1; then
-    echo "Please ensure you can access $__maybe_sudo docker before running this script."
+    echo "Please ensure you can call ${__maybe_sudo:+$__maybe_sudo }docker before running this script."
     exit 1
 fi
 
 if ! cmd --help >/dev/null 2>&1; then
-    echo "Please ensure you can call $__maybe_sudo $__compose_exe before running this script"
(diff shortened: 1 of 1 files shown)


yorickdowne:
This would need to target `main` now.

joeytwiddle:
Understood.

Some things came up. But I will try to get back to this when I can.

yorickdowne:
Thanks for your help! I may take this up once the merge activity dies down a little

PR #894, opened 2022-08-21
Set up docker's host ip

romanoLT:
New based pull request. 
give the ability to specify the Docker's host IP for containers


yorickdowne:
There are a couple things I need to remove ... setting the value and NEW_VARS ... but those are small enough that I can handle them myself. Merging.

PR #954, opened 2022-09-02
Sudo improvements

joeytwiddle:
This completes the `sudo` support first introduced in #760, by doing `sudo` in some more places where it is needed.

(An earlier draft of this PR was in #864, which explains some of the details.)

I tested this branch with a `sudo`-based installation, EL and CL, and it's working for me.

The main changes are in the first commit. The three commits after that are just some housekeeping, and non-critical.

Renaming `cmd()` to `docompose()` made a lot of changes. So it might be easier to read this PR commit-by-commit.

### DIFF
--- a/ethd
+++ b/ethd
@@ -108,6 +108,7 @@ upgrade_compose() {
       echo "Please manually update docker-compose to version 1.29.2. These commands should do it:"
       echo "sudo curl -fsSL "https://github.com/docker/compose/releases/download/1.29.2/docker-compose-$(uname -s)-$(uname -m)" -o /usr/bin/docker-compose"
       echo "sudo chmod +x /usr/bin/docker-compose"
+      exit 1
     fi
   fi
 }
--- a/ethd
+++ b/ethd
@@ -19,7 +19,7 @@ determine_distro() {
     # If Linux, try to determine specific distribution
     if [ "$__uname" == "linux" ]; then
         # If available, use LSB to identify distribution
-        if [ -n "$(which lsb_release)" ]; then
+        if [ -n "$(which lsb_release 2>/dev/null)" ]; then
             __distro=$(lsb_release -i | cut -d: -f2 | sed s/'^\t'//)
         # Otherwise, use release info file
         else
--- a/ethd
+++ b/ethd
@@ -2,7 +2,12 @@
 #set -euo pipefail
 set -uo pipefail
 
+__docker_exe="docker"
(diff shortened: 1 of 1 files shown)


yorickdowne:
OK I'll need you to add something back in. `cmd` is user-facing, so even if it just calls `docompose` with `"$@"`, it needs to exist.

yorickdowne:
Hmm. Why the change to return in upgrade_compose? The intent was that the script exits, so that if there are errors during the process, the user can handle those manually.

ethd
```diff
@@ -834,7 +852,11 @@ restart() {
 }
 
 logs() {
-    cmd logs $@
+    docompose logs $@
+}
+
+cmd() {
+    docompose "$@"
```
joeytwiddle, line 859: I hadn't forgotten to expose `cmd` to the user.

I believe this function provides that.

ethd
```diff
@@ -1448,25 +1470,28 @@ command="$1"
...
 
 determine_distro
+determine_sudo
 handle_root
+determine_docker
 determine_compose
 
 if [ $command = "install" ]; then
     $command "$@"
+    exit "$?"
```
joeytwiddle, line 1480: When I first read:

```
if [ $command = "install" ]; then
    $command "$@"
fi

...
```

I was worried that it didn't exit, and therefore would run `install()` twice!
I had to look inside the `install()` function to find the exit there.

So I moved the `exit` from inside the `install()` function to outside the `install()` function, because I think it's clearer to put that control flow in the script's body.

But this is just a style thing. I don't think it should change the behaviour.

joeytwiddle:
Thanks for this project by the way.

I believe it will keep Ethereum healthy this month!

yorickdowne:
On vacation btw and focusing on merge prep and getting taxes in. I'll get back to this.

ethd
yorickdowne: This means that on ./ethd update it wouldn't exit when upgrading compose. That worries me, there's a risk the update fails and then the user is flooded with error messages from compose on the rest of update.

joeytwiddle: I don't _think_ I removed any `exit` from `upgrade_compose`

But I agree it would be good to have one there!

So I have added an exit on the next commit. Please check if that matches what you think is needed.

yorickdowne: I'll merge this and then see how it behaves.

PR #959, opened 2022-09-04
Make mainnet the default choice

joeytwiddle:
Edited on mobile. I haven't tested this. I'm on vacation.

I guess a few people may start using this project when they see the merge is coming. So it's good to have the wizard offer the most common choice as default.

Do the other questions also default to the most common use-case?

Please feel free to modify this PR, or reproduce it.

### DIFF
--- a/ethd
+++ b/ethd
@@ -841,10 +841,10 @@ query_network() {
     # Mainnet or Testnet network
     NETWORK=$(whiptail --notags --title "Select Network" --menu \
     "Which network do you want to run on?" 12 60 5 \
+    "mainnet" "Ethereum Mainnet" \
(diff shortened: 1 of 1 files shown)


joeytwiddle:
Tested now, and working as expected.

yorickdowne:
I deliberately have testnet as the default. I can move mainnet after Goerli, but don't want to make it the "next next finish" default. Mainnet is scary in that real money is at stake.

yorickdowne:
Closing this in favor of the one I just merged: Mainnet is the second choice. Also removed Ropsten and moved Geth one down: Nethermind is getting there.

PR #962, opened 2022-09-09
use execution as host in prometheus config

casualjim:
The dashboard for erigon didn't fill out any values because it was trying to connect to the wrong hostname


yorickdowne:
Good catch! This changed with Erigon's `alpha`.

PR #973, opened 2022-09-16
use --cache arg for passing GETH_CACHE along

casualjim:
When I tried with a value for GETH_CACHE geth didn't want to come up


yorickdowne:
This would break things. Please look at .env, `GETH_CACHE` isn't just a value.

It could be handled differently in docker-entrypoint.sh but would require careful changes during update so existing setups don't break

PR #983, opened 2022-09-25
lodestar-yaml-quotes

kornface13:
Quotes are needed around the metrics enabled value.


yorickdowne:
Yep they are. Fixed that, thanks!

PR #1048, opened 2022-11-10
Fix check for custom-prom.yml

concentricspheres:
The -f flag is not supported by the shell used, -e is supported.


yorickdowne:
Good find with the path. -f is supported in Posix sh and works; just tested that. Closing in favor of just fixing the path.

PR #1063, opened 2022-11-18
attach-geth command

ldub:
I found myself with a need to attach to geth relatively often and I think it would be nice if there was a convenient command for it. I propose `./ethd attach-geth`, but if you want something more generic that can work with other softwares as well I could try to change it to that. e.g. `./ethd attach execution` and try automatically figure out the correct command for each type of execution client

### DIFF
--- a/ethd
+++ b/ethd
@@ -633,6 +633,20 @@ query_blox_switch() {
     sed -i '/NetworkID/d' blox-ssv-config.yaml
 }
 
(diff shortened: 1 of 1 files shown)


yorickdowne:
Thank you. By all means, if it's useful, though I'd argue that a shell alias for the docker compose command that attaches to Geth would have been just as well :)

PR #1085, opened 2022-12-02
disable early access feature snapshot storage

realsnick:
via issue https://github.com/hyperledger/besu/issues/4764

### DIFF
--- a/besu.yml
+++ b/besu.yml
@@ -74,9 +74,9 @@ services:
       - --metrics-port
       - "6060"
       - --data-storage-format=BONSAI
-      - --Xbonsai-use-snapshots=true
-#      - --sync-mode=X_CHECKPOINT
-      - --sync-mode=X_SNAP
+      - --Xbonsai-use-snapshots=false
+      - --sync-mode=X_CHECKPOINT
+#      - --sync-mode=X_SNAP
 volumes:
   besu-eth1-data:
   jwtsecret:
--- a/besu.yml
+++ b/besu.yml
@@ -74,9 +74,9 @@ services:
       - --metrics-port
       - "6060"
       - --data-storage-format=BONSAI
-      - --Xbonsai-use-snapshots=false
-      - --sync-mode=X_CHECKPOINT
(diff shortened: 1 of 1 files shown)


besu.yml
```diff
@@ -74,9 +74,9 @@ services:
       - --metrics-port
       - "6060"
       - --data-storage-format=BONSAI
-      - --Xbonsai-use-snapshots=true
-#      - --sync-mode=X_CHECKPOINT
-      - --sync-mode=X_SNAP
+      - --Xbonsai-use-snapshots=false
+      - --sync-mode=X_CHECKPOINT
```
macfarla, line 78: did you mean to also change the sync-mode?

realsnick: yes, snapshotting is still alpha

realsnick: my node is almost sync, i'll report if this fixed the issue. I believe it did but let me 100% make sure

macfarla: 👍

realsnick: I can confirm this fixed the issue

realsnick (commented):
works

yorickdowne:
Checkpoint sync is plain worse than snap sync on stability.

As for bonsai snapshots, damned if you do damned if you don't: They were enabled because of https://github.com/hyperledger/besu/issues/4553

I'd like you to get the gitpoap for this one. If you'll just comment out bonsai snapshots and leave the rest as-is, I'll approve.

realsnick:
Question is if snap sync is alpha or not as for me it was not able to recover and dropped my validator below 50% uptime until I resolved it with checkpoint . 

Maybe will be good idea to add this as an option for default.env?

yorickdowne:
Bonsai snapshots certainly have issues. Snap sync itself, the only issue I am aware of is that sometimes it doesn't finish initial sync.

If checkpoint works better for you, I'd say make that change locally and `git commit` it so it stays.

Ultimately, Besu needs to stabilize its sync.

PR #1087, opened 2022-12-02
fix for issue 1086: building nethermind from source fails when checking out submodules

nu404040:
Using --recursive instead of individual submodules to checkout nethermind source.
Put bash -c arguments on multiple lines for readability.


yorickdowne:
Good change. Thank you!

PR #1099, opened 2022-12-13
Enable Teku validator lockfiles

alepacheco:
Teku has a safety feature to lock validator keys when used. This has been disabled but it's enabled by default. 
Unless there is a requirement we should enable it to prevent slashes


robson26:
Is there a reason we shouldn't be merging this in?

alepacheco:
If a user has a hard restart the lockfile won't get removed and it will be hard for most users to manually do it. For that reason I am not sure if it is worth merging

robson26:
Got it, thank you.

PR #1100, opened 2022-12-13
Nethermind not reporting metrics

alepacheco:
The delay has been set really long making nethermind not send any metrics. 
If this is not a requirement I suggest leaving the default


yorickdowne:
This is by design. This keeps Nethermind from sending metrics; instead metrics are polled by Prometheus every scrape interval. Closing.

PR #1102, opened 2022-12-14
Fix Teku SNI

alepacheco:
New teku version requires the hostname be in the file. localhost is not a valid one


0xBaconPancakes:
How can I regenerate the keystore with this fix? I have a running instance, but the keys are already generated with the wrong hostname.

0xBaconPancakes:
Running bash in `docker exec -it eth-docker-consensus-1 bash` and then manually executing all the commands in the paragraph manually works.

yorickdowne:
This unfortunately doesn't quite do it, as it needs to work for validator and consensus containers. I have a fix that I am pushing now.

PR #1110, opened 2022-12-20
auto-prune can handle Nethermind and different docker dir

yorickdowne:


auto-prune.sh
```diff
@@ -1,36 +1,86 @@
...
 fi
 
 cd "$(dirname "$0")" || exit 1
 
-dryrun=0
-for var in "$@"; do
+
+__docker_dir="/var/lib/docker"
+if [ -f "/etc/docker/daemon.json" ]; then
+  __dir=$(jq '."data-root"' /etc/docker/daemon.json)
```
mLewisLogic, line 15: Thanks for the change

We need to make sure we use the `--raw-output` flag for jq, otherwise it will include the double-quotes in the env variable's value itself.

```suggestion
  __dir=$(jq --raw-output '."data-root"' /etc/docker/daemon.json)
```

This change works well in my case, but it will fail for folks who manually change their `data-dir` using the `-g` flag in their systemctl launch config (rather than using `/etc/docker/daemon.json`). Pretty minor issue though and probably not high-pri.

yorickdowne: Thanks for the suggestion! Change made. I am not too worried about users of `-g`.

PR #1119, opened 2022-12-23
Correctly detect docker root dir on rootless docker or -g

yorickdowne:
There's probably a clever go template way of doing this without `tr -d '"'` but I haven't managed to figure that out.

@mLewisLogic this impacts you, different way of detecting docker dir


mLewisLogic:
Oh, that's a brilliant solution! I didn't know about the `--format` option for system info.

PR #1165, opened 2023-02-08
Add option to omit graffiti to use validator default

nflaig:
**Current behavior**

At the moment it is not possible to omit the graffiti flag, even if `GRAFFITI` in the env file is removed or commented out it will add the `--graffiti` flag which results in an empty graffiti being used.

It is not possible to let the validator use the default graffiti which is a lot of times desired if you want to have the client name and version as graffiti since most validators use the `<client-name>-<version>/<short-commit-sha>` as default graffiti.

**Expected behavior**

setting different `GRAFFITI` values in `.env` should behave like this

- `GRAFFITI=🐼eth-docker🐼` should use `🐼eth-docker🐼` as graffiti
- `GRAFFITI=` or `GRAFFITI=""` should result in an empty graffiti
- `GRAFFITI` env removed or commented out should result in omitting `--graffiti` flag and use validator default graffiti

**Proposed change**

Move setting graffiti flag to docker-entrypoint scripts to add additional logic to check if `GRAFFITI` is set


yorickdowne:
`if [ -v VAR ]` is a bash-ism, is it not? Should this be `if [[ -v VAR ]]` ?

nflaig:
As far as I tested this works but needs at least bash version 4.2 which should be the case for all images ([see comment](https://unix.stackexchange.com/questions/212183/how-do-i-check-if-a-variable-exists-in-an-if-statement/212192#comment359524_212192)).

nflaig:
@yorickdowne noticed there is still a minor issue, in the `.env` I added `comment out line below to use default validator graffiti` but the problem with that is that `./ethd update` updates that line if it is either commented out or removed, not really sure what a good solution for this is 🤔 

```
List of changes made to .env during migration - current on left, original on right:

GRAFFITI=                                                     | #GRAFFITI=
```

PR #1173, opened 2023-02-08
Revert graffiti changes

nflaig:
@yorickdowne sorry for the noise but I lost confidence in the changes I did regarding the graffiti

found few potential issues

- commenting out does not work, there are incompatibilities with how eth-docker works currently, see [comment](https://github.com/eth-educators/eth-docker/pull/1165#issuecomment-1423274819)
- current implementation would not work if graffiti is commented out becasue docker-compose automatically set a blank string `WARN[0000] The "GRAFFITI" variable is not set. Defaulting to a blank string.`, this could be fixed by changing `- GRAFFITI=${GRAFFITI}` to just `- GRAFFITI` in `environment`
- graffiti values with spaces also would currently casue issues, as graffiti string is not properly escaped by quotes

I am currently setting up a proper eth-docker test stack to better test my changes in the future 🙏


yorickdowne:
Yep fair. I was a little fast with the merge button as well 😅

yorickdowne:
The issue with docker-compose setting an empty string if it's missing is not easy to solve and is likely your biggest hurdle. Users can have a legitimate reason to want no graffiti.

Alright so spit-balling. This is a corner case. There could be an .env var - like we need any more - called `DEFAULT_GRAFFITI` or something, and if set to `true`, it'll not set graffiti. It and `GRAFFITI` would need to be passed in as env vars, then `GRAFFITI` properly quoted can survive spaces and `DEFAULT_GRAFFITI` would cause it to not be set at all.

nflaig:
I need to understand eth-docker a bit better, maybe there is a leaner solution to achieve this, especially need to dig into how the `ethd` works.

I think in general being able to use the default graffiti of the validator is not that big of a deal. It just makes it a bit easier to manage if you want to use  `<client-name>-<version>/<short-commit-sha>` as graffiti, right now I am just setting this manually to the correct value which is also fine, no big deal tbh

Thanks a lot for your patience 🙏

nflaig:
> Alright so spit-balling. This is a corner case. There could be an .env var - like we need any more - called DEFAULT_GRAFFITI or something, and if set to true, it'll not set graffiti. It and GRAFFITI would need to be passed in as env vars, then GRAFFITI properly quoted can survive spaces and DEFAULT_GRAFFITI would cause it to not be set at all.

thats sound like a good solution

yorickdowne:
`./ethd` is a wrapper around `docker compose`. It started as purely that, and now it also does some other stuff like migrate `.env` variables for better UX, and prep non-interactive key imports, help with pruning geth and nethermind, and such. But at its core - still just a faster way to do `docker compose` stuff.

PR #1174, opened 2023-02-08
Updating docs for setting withdrawal credentials

valefar-on-discord:
Updating documentation to try to differentiate more between the two USB sticks and to recommend creating an additional file that contains the address instead of having the user type manually.


yorickdowne:
This is so much clearer already! Thank you!

PR #1176, opened 2023-02-10
Removing jq as depedency for withdrawal change

valefar-on-discord:
Shell scripting is far from a strength but I wanted to hack together a solution to remove jq as a dependency


yorickdowne:
Oh interesting! Let me grab this PR and try it locally first.

PR #1193, opened 2023-02-19
Add support for Lodestar beaconcha.in monitoring

nflaig:
Hey Yorick 👋 

Lodestar will support beaconcha.in monitoring with `v1.6.0` which is not released yet but the `next` image already supports it.

Opening this as a draft for initial review, I am not sure if we can merge this yet as it might be that a user running Lodestar has `BEACON_STATS_API` which would set `--monitoring.endpoint` CLI flag and cause an error during startup as it is an unknown option to older Lodestar version.

I saw that in `ethd` there would be the possibility to apply some logic as it is done in [this commit](https://github.com/eth-educators/eth-docker/commit/719c8611b46b8e20ef2b3cd95f033f596dd973aa). Here the `BEACON_STATS_API` for example could be removed if Lodestar is below `v1.6.0`. But this might be hard to detect since `latest` tag is usually used by default but possible if a user specified a custom version.

I think best it to just wait for the official `v1.6.0` release and then assume a user will do `./ethd update` and hope they didn't specify a fixed version of Lodestar. Let me know if this would be fine, else can apply some custom logic.


yorickdowne:
This looks great! Yes let's wait until 1.5.0 is out and then merge.

Thanks for all the contributions to eth-docker btw, it's appreciated!

nflaig:
We finally released [v1.6.0](https://github.com/ChainSafe/lodestar/releases/tag/v1.6.0) and support monitoring in the stable release.

PR #1195, opened 2023-02-21
Hide MEV option for RPC node

jimmygchen:
@yorickdowne 
Not sure if `mev-boost` and `coinbase` do anything for an RPC node - I think these options can be hidden?


jimmygchen:
Hey thanks for the explanation, it all makes sense! I knew I was missing something ha, forgot about running the beacon node for remote VCs :) and I agree with you the UX is better this way, so happy to close this.

PR #1197, opened 2023-02-23
Lighthouse jemalloc, Prysm change messages, web3signer

yorickdowne:


yorickdowne:
Lighthouse source build uses jemalloc and builds with slasher support
Prysm defaults changed so credential change messages can be sent via REST
web3signer.yml added - very early and experimental support

PR #1200, opened 2023-03-03
Pin eth-metrics version

BriungRi:
Latest eth-metrics fails with the following error:

```
/usr/local/bin/docker-entrypoint.sh: line 11: /exporter: No such file or directory
```


nflaig:
but if we pin this version of the PR and merge the other PR is will be broken again 😅

BriungRi:
No problem. This can also serve as a reference for other operators who run into this issue.

PR #1202, opened 2023-03-04
ethereum-metrics-exporter fails to start 

alepacheco:
I had this issue where ethereum-metrics-exporter would not start and it was due to this name been different. 
I don't know much about this code but maybe it broke because an update to the container.


lebrauncom:
+1

@alepacheco thanks for finding a solution to this issue. I confirmed your suggested fix does work.

lebrauncom:
Duplicate of #1199

Issue #1217, opened 2023-03-21
Error checking disk space when running ethd update


jimmygchen:
I just ran `ethd update` and received a panic from docker and bash failure in `check_disk_space`.

1. It looks like this command is giving me the panic
```
__docker_dir=$(docker system info --format '{{.DockerRootDir}}')
```
if I run the above command, it gives me the same panic as below. But it works if I run it with `sudo`.

2. and this command returns multiple numbers for me, causing the compare expression to fail:
```
__free_space=$(df -P ${__docker_dir} | awk '/[0-9]%/{print $(NF-2)}')
```

## Environment

OS: Ubuntu 22.04.2 LTS
Docker: 20.10.12

## Full error

```
panic: reflect: indirection through nil pointer to embedded struct [recovered]
	panic: reflect: indirection through nil pointer to embedded struct

goroutine 1 [running]:
text/template.errRecover(0xc0004bd2d8)
	/usr/lib/go-1.17/src/text/template/exec.go:169 +0x145
panic({0x562ad3996aa0, 0x562ad3c523d8})
	/usr/lib/go-1.17/src/runtime/panic.go:1038 +0x215
reflect.Value.FieldByIndex({0x562ad3b1a080, 0xc000122280, 0x0}, {0xc000057220, 0x2, 0x562ad3996aa0})
	/usr/lib/go-1.17/src/reflect/value.go:1224 +0x1e7
text/template.(*state).evalField(0xc0004bd308, {0x562ad3b1a080, 0xc000122280, 0x562ad28b800a}, {0x7fff4d45d6bb, 0xd}, {0x562ad3cbd2b0, 0xc00011ecf0}, {0xc00043ae10, 0x1, ...}, ...)
	/usr/lib/go-1.17/src/text/template/exec.go:616 +0x450
text/template.(*state).evalFieldChain(0x7fff4d45d6ba, {0x562ad3b1a080, 0xc000122280, 0x14}, {0x562ad3b1a080, 0xc000122280, 0x0}, {0x562ad3cbd2b0, 0xc00011ecf0}, {0xc00043ae00, ...}, ...)
	/usr/lib/go-1.17/src/text/template/exec.go:569 +0x1eb
text/template.(*state).evalFieldNode(0x203000, {0x562ad3b1a080, 0xc000122280, 0xc000743380}, 0xc0004bd030, {0xc00043ae10, 0x0, 0x1}, {0x562ad3a1e980, 0x562ad4be6840, ...})
[...]

PR #1218, opened 2023-03-21
Use dodocker to get system info

jimmygchen:
Addresses #1217


yorickdowne:
Thanks! You beat me by half an hour but I only saw it after pushing mine. I'm also checking that awk returns a number, now, just in case.

PR #1224, opened 2023-03-25
Support client default Graffiti

yorickdowne:


lodestar/docker-entrypoint-vc.sh
```diff
@@ -30,6 +30,13 @@ else
   __doppel=""
 fi
 
+# Check whether we should use default graffiti
+if [ "${DEFAULT_GRAFFITI}" = "true" ]; then
+  __graffiti=""
+else
+  __graffiti="--graffiti ${GRAFFITI}"
```
nflaig, line 37: this need to be properly quoted else will run into issues if graffiti contains spaces

Quoting it like this seems to work

```diff
diff --git a/lodestar/docker-entrypoint-vc.sh b/lodestar/docker-entrypoint-vc.sh
index db33b04..1ac7e18 100755
--- a/lodestar/docker-entrypoint-vc.sh
+++ b/lodestar/docker-entrypoint-vc.sh
@@ -34,7 +34,7 @@ fi
 if [ "${DEFAULT_GRAFFITI}" = "true" ]; then
   __graffiti=""
 else
-  __graffiti="--graffiti ${GRAFFITI}"
+  __graffiti='--graffiti "${GRAFFITI}"'
 fi
```

interestingly enough, doing something like `__graffiti="--graffiti \"${GRAFFITI}\""` does not work

nflaig (commented):
@yorickdowne  amazing that you added this! I actually still had that on my list as well ✅

PR #1228, opened 2023-03-25
Fix Graffiti with spaces

yorickdowne:


lodestar/docker-entrypoint-vc.sh
```diff
@@ -32,11 +32,13 @@ fi
 
 # Check whether we should use default graffiti
 if [ "${DEFAULT_GRAFFITI}" = "true" ]; then
+  __graffiti_cmd=""
   __graffiti=""
 else
-  __graffiti='--graffiti "${GRAFFITI}"'
```
nflaig, line 37: ah good catch! I actually only tested if the validator client starts but not the actual command. Might be worth to consider printing out the graffiti on startup in Lodestar, right now we only do it if a block is proposed.

PR #1229, opened 2023-03-25
Truly fix graffiti this time

yorickdowne:


lodestar/docker-entrypoint-vc.sh
```diff
@@ -30,15 +30,12 @@ else
...
+  exec "$@" ${__mev_boost} ${__beacon_stats} ${__doppel} ${VC_EXTRAS}
 else
-  __graffiti_cmd="--graffiti"
-  __graffiti="${GRAFFITI}"
-fi
-
 # Word splitting is desired for the command line parameters
 # shellcheck disable=SC2086
-exec "$@" ${__graffiti_cmd} "${__graffiti}" ${__mev_boost} ${__beacon_stats} ${__doppel} ${VC_EXTRAS}
+  exec "$@" "--graffiti" "${GRAFFITI}" ${__mev_boost} ${__beacon_stats} ${__doppel} ${VC_EXTRAS}
```
nflaig, line 40: now I am curious, what was actually broken in the previous version? I tested it with Lodestar and it looked good to me

yorickdowne: It passed ${GRAFFITI} verbatim to the client. Some clients would maybe read that as an environment var, and other would have a graffiti that literally read ${GRAFFITI}.

So I needed something else. But empty quoted parameters also break clients. Ultimately the only way I found was to violate DRY a little and have two exec lines, one with and one without graffiti.

yorickdowne: Default graffiti has been a rabbit hole. Who'd think such a simple feature would take that much iteration ...

yorickdowne: BTW if Lodestar treats a literal `--graffiti ${GRAFFITI}` line as an env var, that's arguably a bug. The user said the graffiti should be `${GRAFFITI}`. Maybe they're geeky that way.

nflaig: > BTW if Lodestar treats a literal --graffiti ${GRAFFITI} line as an env var, that's arguably a bug

Interesting, will take a look at that, thanks for reporting 👍

nflaig: > BTW if Lodestar treats a literal --graffiti ${GRAFFITI} line as an env var, that's arguably a bug.

I looked into this but if graffiti is set like this `--graffiti '${GRAFFITI}'` Lodestar will use `${GRAFFITI}` as graffiti, of course if I remove the quotes it will the value of `GRAFFITI` env var but this is just bash doing its thing. Should be the case for all other CLs as well

PR #1295, opened 2023-04-13
Fix resync-consensus for lodestart and prysm.

FloatingUpstream:
cl client variable is undefined otherwise


yorickdowne:
Thanks for catching this!

PR #1296, opened 2023-04-13
Fix docker registry for prism validator and beacon-chain image. 

FloatingUpstream:
https://docs.prylabs.network/docs/install/install-with-docker only lists gcr.io as source, 

This https://github.com/prysmaticlabs/prysm/pull/11250/ might have stopped image updates on docker.io. Consequently e.g. stable now points to an outdated image on docker.io, no idea who updates those.


yorickdowne:
They still update on docker hub, see for example https://hub.docker.com/r/prysmaticlabs/prysm-beacon-chain/tags

However `stable` on docker hub is behind `stable` on gcr.io. 

Looks like a good change to make, thank you!

Issue #1336, opened 2023-04-24
Validator Exit for Lighthouse failing


gorillamania:
I'm trying to exit one of my validator nodes with the following command (as described on the [documentation](https://eth-docker.net/Support/Exit/#lighthouse))

```bash
➜ ./ethd cmd run --rm validator-exit /keys/keystore-m_12381_3600_1_0_0-1679192309.json
cp: -r not specified; omitting directory '/validator_keys/keybackup'
```
With no other output.

I noticed that if I put a random filename in, it does the same behavior:
```
➜ ./ethd cmd run --rm validator-exit file_not_there                                   
cp: -r not specified; omitting directory '/validator_keys/keybackup'
```

My `.eth/validator_keys` directory does indeed contain that file. 

Of note the instructions say to use `/keys/$keystore_filename`, but when I connect to the docker container, the `/keys` directory does not exist. 

```╰─❯ docker exec -it -u root eth-docker-validator-1 bash
root@4d2fa1db71d8:/# ls -l
total 56
lrwxrwxrwx   1 root root    7 Mar  8 02:08 bin -> usr/bin
drwxr-xr-x   2 root root 4096 Apr 18  2022 boot
drwxr-xr-x   5 root root  340 Apr 23 23:49 dev
drwxr-xr-x   1 root root 4096 Apr 23 23:49 etc
drwxr-xr-x   2 root root 4096 Apr 18  2022 home
lrwxrwxrwx   1 root root    7 Mar  8 02:08 lib -> usr/lib
drwxr-xr-x   2 root root 4096 Mar  8 02:08 media
drwxr-xr-x   2 root root 4096 Mar  8 02:08 mnt
drwxr-xr-x   2 root root 4096 Mar  8 02:08 opt
dr-xr-xr-x 239 root root    0 Apr 23 23:49 proc
drwx------   1 root root 4096 Apr 24 00:11 root
drwxr-xr-x   5 root root 4096 Mar  8 02:13 run
lrwxrwxrwx   1 root root    8 Mar  8 02:08 sbin -> usr/sbin
drwxr-xr-x   2 root root 4096 Mar  8 02:08 srv
dr-xr-xr-x  13 root root    0 Apr 23 23:49 sys
drwxrwxrwt   1 root root 4096 Apr 23 23:49 tmp
drwxr-xr-x   1 root root 4096 Mar  8 02:08 usr
drwxr-xr-x   1 root root 4096 Mar  8 02:12 var
[...]

PR #1337, opened 2023-04-24
Fix for Lighthouse validator exit if there is a keybackup directory

gorillamania:
Issue https://github.com/eth-educators/eth-docker/issues/1336


yorickdowne:
Thanks! That'll work. Copying `*.json` would maybe be another way, doing a find -type f an even more granular one. Since `/keys` is gone after, I think this is fine.

PR #1353, opened 2023-04-30
Increase torrent download speed during initial sync

alepacheco:
The default speed limit is 16mb/s resulting in around 4/5h of download time.
With this change it should be <1h depending on the connection speed.


yorickdowne:
On reflection I do not wish to set this for everyone. People don't necessarily have that kind of speed. You can place this in `EL_EXTRAS` inside `.env` and it'll work that way for you.

PR #1397, opened 2023-05-15
Add prometheus-shared.yml to expose prometheus port

allen-pattern:
Along the lines of all the other `*-shared.yml` configuration files. I'm using this locally for federation.


yorickdowne:
This won't work as-is, because Prometheus itself doesn't use `PROMETHEUS_PORT` and so would still run on 9090.

allen-pattern:
Thanks. I've changed it to only affect the exposed port.

Issue #1433, opened 2023-06-21
Building nethermind from source with default NM_SRC_BUILD_TARGET may get source with "unstable" tag.


nu404040:
Building nethermind from source with NM_SRC_BUILD_TARGET from default.env may result in checking out from a git tag that specifies work in progress - not an actual release.   For example, at the time of creation of this issue, tag "1.20.0-rc" (release candidate) will be used but "1.19.2" would  be the appropriate tag to use.

yorickdowne:
PR accepted

PR #1434, opened 2023-06-21
Using grep to limit default NM_SRC_BUILD_TARGET 

nu404040:
Using grep to limit default NM_SRC_BUILD_TARGET to tags that match
regex: "^[0-9]+[.][0-9]+[.][0-9]+$" .  This avoids building from tags like "1.20.0-rc" or "1.17.0-unstable"

This is a fix for issue  #1433


yorickdowne:
My assumption was that someone who source-builds will seek out a specific tag, and the default is just there as a fallback - and yes, this seems fine to change.

PR #1447, opened 2023-07-13
feat: add reth to ethd config setup

d1onys1us:


d1onys1us:
closing for now as discussed to be alpha software. however it's still very easy to add reth.yml in `.env`.

PR #1454, opened 2023-07-24
feat: add grafana-auth.yml

d1onys1us:
## Reason for PR
i often run multiple instances of the eth-docker (and subsequently, grafana, in a single browser window). However, grafana adds some browser storage value for auth, and personally, i like to just disable auth here. so adding the ability to view the dashboard without auth is a nice feature to have for me so I'm thinking it will be helpful for others too.

### DIFF
--- a/grafana-auth.yml
+++ b/grafana-auth.yml
@@ -1,0 +1,5 @@
+version: "3.9"
+services:
+  grafana:
+    environment:
+      - GF_AUTH_ANONYMOUS_ENABLED=true


yorickdowne:
I am inclined not to add this to the repo. Disabling auth on Grafana is a security worry; people who want this can always add it themselves and commit the change; and grafana-cloud.yml is designed to send data to a central Mimir, Loki, Grafana stack, whether that's self-hosted or the actual Grafana Cloud.

yorickdowne:
Another way to achieve this, btw, is to `sudo bash` into the `grafana-config` volume and change `grafana.ini` the way you want it.

Bind-mounting this volume is a little tricky with permissions. Grafana runs as user `grafana` id 472, and its directories are owned by `grafana:root`. That's why it's a named volume. You can still get in there and change the contents of `grafana.ini`

PR #1505, opened 2023-09-07
Remove the default http.api configuration to avoid overwriting configuration in geth.yml.

jiangbo0216:
https://geth.ethereum.org/docs/interacting-with-geth/rpc#http-server

The default whitelist allows access to the eth, net and web3 namespaces.

### DIFF
--- a/geth/docker-entrypoint.sh
+++ b/geth/docker-entrypoint.sh
@@ -51,7 +51,7 @@ if [[ "${NETWORK}" =~ ^https?:// ]]; then
     touch /var/lib/goethereum/setupdone
   fi
 else
-  __network="--${NETWORK} --http.api web3,eth,net"
+  __network="--${NETWORK}"
 fi
 
 # Set verbosity


yorickdowne:
What are you looking to accomplish? This used to be in geth.yml and is now in docker-entrypoint.sh so it works with the custom testnet setup

yorickdowne:
EL_EXTRAS is the usual way to add parameters to the EL. This is less intrusive than editing the yml and committing the change.

The api parameter moved into the entrypoint to support custom testnets, where a broader set of APIs is enabled by default. It’s debatable whether that was strictly necessary. As long as specifying the parameter with EL_EXTRAS again works, I am inclined to leave it as is.

If EL_EXTRAS cannot override the parameter then I think you are right, another solution would be necessary.

I will think on these defaults some. I am trying to avoid a situation where the defaults change and a broader API range is enabled without that being a deliberate choice.

jiangbo0216:
get your point，I agree that to have a lot available in .env is better.

for these defaults,  if eth-docker's geth component is same with the official default is better, it can reduce cognitive load.

and I understand your consideration,  avoid a situation where the defaults change and a broader API range is enabled is ok, but if there are difference between eth-docker and official geth. it need to document. 

I appreciate your input and thoughtful contribution. I offer my ideas, I hope it will be helpful.

If everything is clear, we can close this pull request.

yorickdowne:
Fair enough, you convinced me

PR #1569, opened 2023-10-18
Prysm rest

### LINKED ISSUE #1563: Prysm vc only always has builder enabled / client interop
I am using eth-docker to do interop testing between Lodestar and other clients. For that I am using the vc only files which works great and makes interop testing really smooth but I noticed the `prysm-vc-only.yml` always has builder enabled and there is no option to enable the REST API which is required to work with other clients.

https://github.com/eth-educators/eth-docker/blob/38982d69e22b669d4aeb45ac0e1a949dfb43e842/prysm-vc-only.yml#L71

Wondering if it makes sense to support a interop mode for prysm, although I can't tell if anyone besides me would actually use that. If not I think this issue can be closed by just removing the builder flag as it is already set in the entrypoint script

https://github.com/eth-educators/eth-docker/blob/38982d69e22b669d4aeb45ac0e1a949dfb43e842/prysm/docker-entrypoint-vc.sh#L33-L35

Just dropping my git diff here to make it work
```diff
diff --git a/prysm-vc-only.yml b/prysm-vc-only.yml
index 80dfc02..f5650eb 100644
--- a/prysm-vc-only.yml
+++ b/prysm-vc-only.yml
@@ -61,13 +61,15 @@ services:
       - --grpc-gateway-port
       - ${KEY_API_PORT:-7500}
       - --grpc-gateway-corsdomain=*
-      - --beacon-rpc-gateway-provider
-      - consensus:5052
+#      - --beacon-rpc-gateway-provider
+#      - consensus:5052
       - --suggested-fee-recipient
       - ${FEE_RECIPIENT}
-      - --enable-builder
       - --wallet-password-file
       - /var/lib/prysm/password.txt
+      - --beacon-rest-api-provider
+      - ${CL_NODE}
+      - --enable-beacon-rest-api
     labels:
       - traefik.enable=true
       - traefik.http.routers.prysm.entrypoints=web,websecure
@@ -140,8 +142,8 @@ services:
       - --${NETWORK}
       - --backup-dir=/validator_keys
       - --backup-password-file=/var/lib/prysm/password.txt
-    depends_on:
-      - consensus
+#    depends_on:
+#      - consensus

   validator-keys:
     profiles: ["tools"]
```

yorickdowne:
Change Prysm VC to use REST. Prysm CL no longer exposes gRPC.

Remove legacy exit tooling, as prysmctl only supports gRPC

Warn user during migration

Merge after Prysm REST is considered fully stable under Gloas and reorgs

`default.env` version needs to be adjusted before merge

Closes #1563 

See https://github.com/prysmaticlabs/prysm/issues/13519

### DIFF
--- a/prysm-vc-only.yml
+++ b/prysm-vc-only.yml
@@ -62,8 +62,6 @@ services:
       - /var/lib/prysm/
       - --keymanager-token-file
       - /var/lib/prysm/auth-token
-      - --beacon-rpc-provider
-      - ${CL_NODE:-consensus:4000}
       - --verbosity
       - ${LOG_LEVEL}
       - --accept-terms-of-use
@@ -76,6 +74,8 @@ services:
       - --http-port
       - ${KEY_API_PORT:-7500}
       - --rpc
+      - --beacon-rest-api-provider
+      - ${CL_NODE:-http://consensus:5052}
       - --suggested-fee-recipient
       - ${FEE_RECIPIENT}
       - --wallet-password-file
--- a/prysm-vc-only.yml
+++ b/prysm-vc-only.yml
@@ -62,8 +62,6 @@ services:
       - /var/lib/prysm/
       - --keymanager-token-file
       - /var/lib/prysm/auth-token
-      - --beacon-rpc-provider
-      - ${CL_NODE:-consensus:4000}
       - --verbosity
       - ${LOG_LEVEL}
       - --accept-terms-of-use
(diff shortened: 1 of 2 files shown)


yorickdowne:
gRPC is deprecated in v6.0.0 but REST isn't recommended yet for the validator. Possibly with v8 in 2026, it's expected Prysm will remove gRPC

lucylou17 (approved):

lucylou17 (approved):

yorickdowne:
Prysm 8.0.0 makes REST a first-class citizen, if not the default for the VC. Expected in time for Glamsterdam hardfork.

Edit: Nope I jumped the gun on that, I've been set straight. Hang loose.

PR #1603, opened 2023-11-04
Remove pre-merge legacy code, remove compose V1 code

yorickdowne:
Merge this after Dencun


yorickdowne:
There needs to be a check for the ERIGON_DOCKER_TAG before prompting users to delete the DB. They may stay on v2

gathecageorge (approved):
Everything looks ok

PR #1621, opened 2023-11-11
Metric collection compatibility enhancement

aliask:
### Change overview
- Increased default polling rate to 15s
- Manually reduced the price-check poll rates to 5m, no point hammering the APIs
- Allow override of job name and polling interval via Docker labels

### Explanation
Eth-docker currently does not set the polling configs, so it uses the default values (1 minute interval).

Several publicly shared Grafana dashboards expect Prometheus metrics to be available at a higher rate than what `eth-docker` provides, so importing these dashboards into Grafana causes some widgets to display "No Data".

These values were taken from the [Rocketpool config](https://github.com/rocket-pool/smartnode-install/blob/master/install/prometheus.tmpl), but the official [lighthouse-metrics](https://github.com/sigp/lighthouse-metrics/blob/master/prometheus/prometheus.yml) config also uses 15s, and [geth](https://geth.ethereum.org/docs/monitoring/dashboards#setting-up-prometheus) even suggests 10s.

I've also added some extra config overrides to set the job name and poll interval via Docker labels. If these labels are omitted it still defaults to the compose service name.

In my case, this was done for compatibility with the Rocketpool dashboard as they name their jobs differently service name, but there may be other situations where this would be helpful.

I know that the polling settings can be overriden via a `custom-prom.yml` file, however the job name and individual container interval settings cannot (as far as I understand).

The increased polling rate will give users higher resolution metrics as a bonus, and the additional load on resources should not be significant.


aliask:
I've added a few new features:
- Uses `yq` tool to perform a merge of the default and custom configs. [Ref](https://mikefarah.gitbook.io/yq/operators/multiply-merge#merge-appending-arrays)
- Rootless configs now also use a folder, `rootless.d` for their scrape targets
- Configuration is back to being regenerated at each startup instead of at build-time
- A new `.env` setting has been added to allow bypassing config generation, and using `custom-prom.yml` as a standalone configuration file

I'm not so sure about that last one - the other features should make the configuration extensible enough for 99% of users.

prometheus/Dockerfile
```diff
@@ -1,15 +1,13 @@
...
-RUN chown nobody:nobody /etc/prometheus/rootless/*.yml
-USER nobody
-
-ENTRYPOINT make-config.sh
-CMD ["/bin/prometheus", "--storage.tsdb.path=/prometheus", "--web.console.libraries=/usr/share/prometheus/console_libraries", "--web.console.templates=/usr/share/prometheus/consoles"]
+COPY --from=yq /usr/bin/yq /usr/local/bin/yq
+
+COPY base-config.yml rootless-base-config.yml ./
+COPY ./rootless/*.yml ./rootless/
+COPY ./docker-entrypoint.sh /
```
yorickdowne, line 11: Into /usr/local/bin/ to match what other locally-built containers do?

prometheus/docker-entrypoint.sh
```diff
@@ -0,0 +1,87 @@
...
+for var in "$@"; do
+  case "$var" in
+    --config.file* )
+      echo "WARNING - Manual setting of --config.file found, bypassing automated config preparation in favour of supplied argument"
+      /bin/prometheus "$@"
+  esac
+done
+
+prepare_config
+/bin/prometheus "$@" --config.file="${__config_file}"
```
yorickdowne, line 87: Add an `exec` here so `/bin/prometheus` receives `SIGINT` and `SIGTERM`

prometheus/docker-entrypoint.sh
```diff
@@ -0,0 +1,87 @@
...
+    echo "No Client information passed, running in Docker target detection mode"
+    __base_config=base-config.yml
+  else
+    echo "Client information detected, compiling prometheus config"
+    select_clients
+    __base_config=rootless-base-config.yml
+  fi
+
+  # Merge custom config overrides, if provided
+  if [ -s custom-prom.yml ]; then
```
yorickdowne, line 65: This gets me `No custom configuration detected` even when there is one. Likely needs to be `if [ -s /etc/prometheus/custom-prom.yml ]` and then likewise use that path for the `yq` further down.

yorickdowne (changes requested):

prometheus/docker-entrypoint.sh
aliask: I've updated the script to `cd /etc/prometheus` at the top, and used all relative paths for the preparation steps.

I've kept it so that Prometheus gets started with the absolute path for the prepared config file.

Tested this working with `grafana.yml` (not rootless) and it's looking OK here.

yorickdowne (approved):

yorickdowne:
Works well here, both my global external labels as well as the remote write are preserved. Merging. Thank you!

PR #1694, opened 2024-01-04
Enable validator only setup on gnosis chain

haurog:
Separated validator clients from the nodes themselves on my own servers. eth-docker does not support validator only configuration on gnosis chain. I added the 'validator client only' selection to the config screen in gnosis chain. It sets the "validator" string in deployment. All the logic afterwards was already there, so I only had to add it to the config screen.


yorickdowne:
Sure thing. This was mostly used for RocketPool integration through config, which doesn't exist on Gnosis Chain. I figured anyone who runs something more advanced will want to not use config and configure manually instead, particularly since the ext-network.yml isn't useful in this instance. That said, I see no harm. Thanks for the contribution!

PR #1706, opened 2024-01-13
Grandine support

yorickdowne:


sauliusgrigaitis:
@yorickdowne thanks for integrating Grandine. We implemented full Keymanager API except Voluntary Exit, it's available in the latest `unstable` docker images.

We deprectated `web3signer-api-url` (the new name is `web3signer-url`). 

In order to use Keymanager API, `--keystore-storage-password-file` must be passed. 

Could you update the PR accordingly including the option to choose Grandine in the UI? It would be nice to have these changes in order to test it.

yorickdowne:
```
      --keystore-storage-password-file <KEYSTORE_STORAGE_PASSWORD_FILE>
          Path to a file containing password for decrypting imported keystores from API
```

I don't see how that's going to work. I don't want to have to special-case for Grandine and mandate that all keystores have the same password if Grandine is in use.

The `POST` to import local keystore(s) contains the password(s) for those: https://ethereum.github.io/keymanager-APIs/#/Local%20Key%20Manager/importKeystores

The rest of the changes are in.

sauliusgrigaitis:
Hm... This is the single password that protects all the keystores imported via Keymanager API. It's something like local wallet. In this way no individual password of each imported keystore is required to store on disk. Only the new password is needed to secure the combined keystore that contains all the keystores imported via Keymanager API.

yorickdowne:
Thanks for the explanation, that clears it up. That change is in.

sauliusgrigaitis:
@yorickdowne as per TG discussion we added:

```
--enable-validator-api
--validator-api-address <VALIDATOR_API_ADDRESS>
--validator-api-port <VALIDATOR_API_PORT>
--validator-api-allowed-origins <VALIDATOR_API_ALLOWED_ORIGINS>
--validator-api-timeout <VALIDATOR_API_TIMEOUT>
```

The `api-token.txt` is created at `validator/api-token.txt`. These changes are not released, could you updated the PR then I can test with local build?

yorickdowne:
Done, go ahead and test

PR #1714, opened 2024-01-24
Update Lodestar beacon max-old-space-size to 8192

nflaig:
**Motivation**
- https://github.com/ChainSafe/lodestar/pull/6343

**Description**

Update Lodestar beacon max-old-space-size to 8192

### DIFF
--- a/lodestar/Dockerfile.source
+++ b/lodestar/Dockerfile.source
@@ -65,4 +65,6 @@ RUN chmod -R 755 /usr/local/bin/*
 
 USER lsconsensus
 
-ENTRYPOINT ["node", "--max-old-space-size=16384", "/usr/app/node_modules/.bin/lodestar"]
+ENV NODE_OPTIONS=--max-old-space-size=8192
+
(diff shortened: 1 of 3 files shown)


lodestar/Dockerfile.source
```diff
@@ -65,4 +65,6 @@ RUN chmod -R 755 /usr/local/bin/*
 
 USER lsconsensus
 
-ENTRYPOINT ["node", "--max-old-space-size=16384", "/usr/app/node_modules/.bin/lodestar"]
```
nflaig, line 68: I don't think this change makes a difference as entrypoint is always overriden in yaml files but this aligns the Dockerfile closer to what we have in Lodestar, see [Dockerfile#L45](https://github.com/ChainSafe/lodestar/blob/43994ba8ef665c98d7ec6dcbb5338ebf5906f249/Dockerfile#L45)

yorickdowne:
Thank you! This kind of engagement is invaluable to me.

PR #1721, opened 2024-01-29
Review lodestar cli flags

nflaig:
I did a quick review of Lodestar CLI flags

- removed --force as it should no longer be required
- updated deprecated flags

### DIFF
--- a/lodestar-vc-only.yml
+++ b/lodestar-vc-only.yml
@@ -90,9 +90,9 @@ services:
       - voluntary-exit
       - --dataDir
       - /var/lib/lodestar/validators
-      - --server
+      - --beaconNodes
       - ${CL_NODE:-http://consensus:5052}
-      -  --logLevel
+      - --logLevel
       - ${LOG_LEVEL}
       - --network
       - ${NETWORK}
--- a/lodestar-vc-only.yml
+++ b/lodestar-vc-only.yml
@@ -66,7 +66,6 @@ services:
       - ${LOG_LEVEL}
       - --suggestedFeeRecipient
       - ${FEE_RECIPIENT}
-      - --force
     labels:
       - metrics.scrape=true
       - metrics.path=/metrics
@@ -97,7 +96,6 @@ services:
       - ${LOG_LEVEL}
       - --network
       - ${NETWORK}
-      - --force
 
   validator-keys:
     profiles: ["tools"]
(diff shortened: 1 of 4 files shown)


lodestar-vc-only.yml
```diff
@@ -66,7 +66,6 @@ services:
       - ${LOG_LEVEL}
       - --suggestedFeeRecipient
       - ${FEE_RECIPIENT}
-      - --force
```
nflaig, line 69: I haven't checked but I'd assume this flag was added to allow VC to start up after an unexpected shutdown (e.g. due to power loss)

With the next release this won't be required as we improved lockfile handling
- https://github.com/ChainSafe/lodestar/pull/6363

Better to remove this flag

lodestar.yml
```diff
@@ -164,13 +163,12 @@ services:
...
       - /var/lib/lodestar/validators
-      - --server
+      - --beaconNodes
       - http://consensus:5052
-      -  --logLevel
+      - --logLevel
       - ${LOG_LEVEL}
       - --network
       - ${NETWORK}
-      - --force
```
nflaig, line 173: For voluntary exit this is also not required any more, the command can be executed while the VC is running
- https://github.com/ChainSafe/lodestar/pull/5950

lodestar-vc-only.yml
yorickdowne: Yes, it was added to deal with unclean shutdown like a power loss.

yorickdowne:
This will require which Lodestar version?

nflaig:
> This will require which Lodestar version?

It will work with any of the more recent releases but to get the stale lockfile fix in, would have to wait for 1.15.1, we have just cut a rc for this, will be released this week.

Will mark this as ready once the release is out

yorickdowne:
OK. I may want to bundle this with #1682 (Besu 14.1.2) so that I don't have to increment the Eth Docker major twice, as both are breaking changes.

nflaig:
[v1.15.1](https://github.com/ChainSafe/lodestar/releases/tag/v1.15.1) has been released

yorickdowne:
Great! I’d like to hold on to this for a bit, see what ACDC says today about client releases. If I can bundle this breaking change with other breaking changes, that’d be nice. Saves me from needing to rapidly increase the version with every change.

PR #1793, opened 2024-04-05
chore: :arrow_up: bumps Erigon to v2.59.3

victorelec14:
https://github.com/ledgerwatch/erigon/releases/tag/v2.59.3


yorickdowne:
Thank you. Really we need a `stable` tag back - `default.env` is only used when there is no `.env`, and after that it's up to the user to change tags.

victorelec14:
@yorickdowne make sense to use the "devel" tag or is it a bit dangerous?  https://hub.docker.com/r/thorax/erigon/tags

I don't know why Eren doesn't want to use the Latest or Stable tag. Another alternative could be to use the repository directly like Rocketpool or Stader does?

Thanks!

yorickdowne:
I’m not keen on devel. For now users need to manually set the release they want, until Erigon decides to provide a stable or latest tag again.

PR #1798, opened 2024-04-09
Loki3

yorickdowne:
Earliest June 1st 2024 -> v13 has to be active for 30d first


yorickdowne:
Loki will actually handle this cleanly. The only issue might be with users who did not seek an Eth Docker update and now get the v13 schema after May 1st - there isn't a really good central solution to that.

PR #1831, opened 2024-05-14
Migrate to new volume names

yorickdowne:
For `lighthouse.yml` and `prysm.yml`, so users can more easily switch to/from `-cl-only.yml`.

Remove conversion logic no sooner than 1 month after Pectra

### DIFF
--- a/lighthouse/docker-entrypoint.sh
+++ b/lighthouse/docker-entrypoint.sh
@@ -6,6 +6,16 @@ if [ "$(id -u)" = '0' ]; then
   exec gosu lhconsensus docker-entrypoint.sh "$@"
 fi
 
+# Migrate from old to new volume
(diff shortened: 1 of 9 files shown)


lighthouse/docker-entrypoint.sh
```diff
@@ -6,6 +6,16 @@ if [ "$(id -u)" = '0' ]; then
   exec gosu lhconsensus docker-entrypoint.sh "$@"
 fi
 
+# Migrate from old to new volume
+if [[ -d /var/lib/lighthouse-og/beacon && ! -f /var/lib/lighthouse-og/beacon/migrationdone \
+    && $(ls -A /var/lib/lighthouse-og/beacon/) ]]; then
+  echo "Migrating from old Lighthouse volume to new one"
+  echo "This may take 10 minutes on a fast drive. Please be patient"
```
pesho, line 13: Welp, I thought I had a fast SSD, but in my case the migration took about 2 hours

PR #1851, opened 2024-06-12
chore: fix some comments

threehonor:


yorickdowne:
Thank you. TIL that `supercede` while a variant spelling widely in use, is commonly considered an error.

PR #1855, opened 2024-06-18
add network label to services

gathecageorge:
This will add a network label for every service there is metrics.scrape label. This label will be used by prometheus together with the rest when scraping the services for metrics and allow one to differentiate in promQL and/or grafana query between instances of eth-docker running on different networks like sepolia, mainnet, holesky...etc

When this is used together with `https://github.com/CryptoManufaktur-io/central-proxy-docker` then prometheus will add a label network with the value of eth-docker network to all metrics scrapped

### DIFF
--- a/besu.yml
+++ b/besu.yml
@@ -79,6 +79,7 @@ services:
       - metrics.path=/metrics
       - metrics.port=6060
       - metrics.instance=execution
+      - metrics.network=${NETWORK}
 
   set-prune-marker:
     profiles: ["tools"]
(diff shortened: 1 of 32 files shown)


yorickdowne:
Given that this is narrowly useful to ourselves, and the network will be the same throughout a stack: Wouldn't it be easier to make this a custom label in custom-prom.yml so we don't have to change Eth Docker for it?

gathecageorge:
eth-docker allows one to use any network, meaning there might be others also using it for mainnet, sepolia, holesky. When gathering metrics from one or more instances of eth-docker, then it can be hard to know which is which like our case.

Also you can have external prometheus instead of using prometheus within eth-docker, in that case external prometheus wont have access to network variable in environment variables and would require labels.

yorickdowne (approved):
Looks good. Approved

PR #1903, opened 2024-08-12
Switch to use CHECKPOINT sync for besu

antonydenyer:
https://github.com/hyperledger/besu/releases/tag/24.7.1

CHECKPOINT sync use less storage and is faster than snap

https://besu.hyperledger.org/public-networks/get-started/connect/sync-node


yorickdowne:
`CHECKPOINT` sync is not friendly to the network. Other nodes can't sync from a Besu that has been checkpoint synced.

That said, `BONSAI` and `SNAP` are the defaults by now. I can rely on defaults and let users override those via `EL_EXTRAS` if they desire.

Rejected in favor of #1904

PR #1916, opened 2024-08-30
Mev source target

yorickdowne:
Change default target to `stable`

Tested source build with mev-boost `1.8`

### DIFF
--- a/default.env
+++ b/default.env
@@ -176,7 +176,7 @@ SSV_DKG_TAG=latest
 
 # MEV-Boost
 # SRC build target can be a tag, a branch, or a pr as "pr-ID"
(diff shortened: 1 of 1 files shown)


jackvreeken:
Can you add a bit of context "why" in the commit message? Like, is the "stable" tag new, or did we need "develop" for some feature until now, and "stable" has caught up?

pietjepuk2:
Can you add a bit of context "why" in the commit message? Like, is the "stable" tag new, or did we need "develop" for some feature until now, and "stable" has caught up?

yorickdowne:
How's this?

pietjepuk2 (approved):

PR #1920, opened 2024-08-31
Remove Prysm deprecated `--enable-debug-rpc-endpoints`

yorickdowne:
Deprecated with Prysm `v5.1.0`. Removal now means no failures later.


pietjepuk2:
Good to mention it's now on by default, so nothing is expected to break.

pietjepuk2 (approved):

PR #1923, opened 2024-09-09
New Erigon repo and tag

yorickdowne:


yorickdowne:
See https://github.com/erigontech/erigon/issues/11932 as to why this fails CI

PR #1932, opened 2024-09-17
Titan relay

yorickdowne:
For mainnet. Not for SSV, as it's not supported there yet.

https://docs.titanrelay.xyz/


pietjepuk2:
Titan stuff looks good.. can you explain a bit about what the "ejector messages" are? I haven't heard of that term before

yorickdowne:
I confused myself with two PRs … one sec

pietjepuk2 (approved):

PR #1933, opened 2024-09-19
discv5 Geth is now enabled by default

yorickdowne:
And can therefore be removed from the entrypoint script


pietjepuk2 (approved):
LGTM

yorickdowne:
I’ll hold off on merging this until the new Geth is actually on docker hub. They appear to have some CI troubles.

PR #1950, opened 2024-10-03
Fix: Broken `Lido DV Exit` and `Lido validator ejector` when a secondery CL client is added to `Charon`

cnupy:
Adding a secondary CL node endpoint to the `OBOL_CL_NODE` variable breaks the `validator-ejector` and `lido-dv-exit` services because they don't support fallback endpoints.

This PR introduces an additional environment variable `OBOL_CHARON_CL_ENDPOINTS` specific to the `Charon` service, which should be used to set the CL client for the `Charon` service. Multiple endpoints can be separated by commas.


default.env
```diff
@@ -193,6 +195,7 @@ VE_LOCATOR_ADDRESS=
 VE_ORACLE_ADDRESSES_ALLOWLIST=
 ENABLE_DIST_ATTESTATION_AGGR=
 LIDO_DV_EXIT_EXIT_EPOCH=
+LIDO_DV_EXIT_VERSION=
```
yorickdowne, line 198: Where is this variable being used?

default.env
```diff
@@ -165,7 +165,9 @@ EL_NODE=http://execution:8551
 CL_NODE=http://consensus:5052
 # MEV-boost address. This would only be changed for Vouch setups
 MEV_NODE=http://mev-boost:18550
-# Consensus client address for Charon and Lido Validator Ejector in Obol setup
+# Consensus client addresses for Charon in Obol setup
+OBOL_CHARON_CL_ENDPOINTS=http://consensus:5052
```
yorickdowne, line 169: Add this to `__env_migrate` in `ethd` as well, so its value is kept across `ethd update`

yorickdowne: Increment `ENV_VERSION` below from currently `17` to `18`, so an `ethd update` will bring in the new variable you defined.

yorickdowne (changes requested):

default.env
cnupy: > Add this to `__env_migrate` in `ethd` as well, so its value is kept across `ethd update`

Sorry, this is my first post and I have a few questions:
I need to add the variable to the `__all_vars` array to be persisted after the `ethd update` - is this correct?
But I'm not sure if there is a mechanism to initialise the value of the new variable on the first update? I'm thinking that someone who has changed the value of the `OBOL_CL_NODE` variable would want the new `OBOL_CHARON_CL_ENDPOINTS` variable to be set to the value of the `OBOL_CL_NODE` variable when it is introduced. Do you have any guidance on how to implement this?
I also noticed that in the `__all_vars` array, the name of the `OBOL_CL_NODE` variable is incorrectly written as `CL_OBOL_NODE`.

default.env
cnupy: > Where is this variable being used?

It's used in `lido-obol.yml` on line 88, but was missing from the `default.env`

yorickdowne: Oh good catch! Into `__env_migrate` as well then, please.

default.env
yorickdowne: Yes that's correct, it goes into `__all_vars`, which causes it to persist when `ENV_VERSION` is incremented.

Thanks for the typo catch, can you fix that while at it?

There is a mechanism to initialize. Since the variable is new, you could check for `ENV_VERSION` and if below the one you are incrementing to here, set `OBOL_CHARON_CL_ENDPOINTS` to whatever `OBOL_CL_NODE` is. You can find some sample code that does similar things in `__env_migrate`. Well, not merged yet ... take a look at the v6 enable PR.

cnupy: Thanks. All resolved.

ethd
```diff
@@ -1032,16 +1032,16 @@ __env_migrate() {
...
     AUTOPRUNE_NM LOGS_LABEL CF_DNS_API_TOKEN CF_ZONE_API_TOKEN CF_ZONE_ID AWS_PROFILE AWS_HOSTED_ZONE_ID \
     GRAFANA_HOST SIREN_HOST DISTRIBUTED BESU_HEAP TEKU_HEAP PROM_HOST HOST_IP SHARE_IP PRYSM_HOST EE_HOST \
     EL_HOST EL_LB EL_WS_HOST EL_WS_LB CL_HOST CL_LB VC_HOST DDNS_SUBDOMAIN IPV6 DDNS_PROXY RAPID_SYNC_URL \
-    CL_NODE CL_OBOL_NODE BEACON_STATS_API BEACON_STATS_MACHINE EL_P2P_PORT CL_P2P_PORT WEB3SIGNER PRYSM_PORT DOPPELGANGER \
+    CL_NODE OBOL_CL_NODE BEACON_STATS_API BEACON_STATS_MACHINE EL_P2P_PORT CL_P2P_PORT WEB3SIGNER PRYSM_PORT DOPPELGANGER \
     PRYSM_UDP_PORT CL_QUIC_PORT GRAFANA_PORT SIREN_PORT PROMETHEUS_PORT KEY_API_PORT TRAEFIK_WEB_PORT \
     TRAEFIK_WEB_HTTP_PORT CL_REST_PORT EL_RPC_PORT EL_WS_PORT EE_PORT ERIGON_TORRENT_PORT LOG_LEVEL JWT_SECRET \
     EL_EXTRAS CL_EXTRAS VC_EXTRAS ARCHIVE_NODE SSV_P2P_PORT SSV_P2P_PORT_UDP OBOL_P2P_PORT ERIGON_P2P_PORT_2 \
-    ERIGON_P2P_PORT_3 LODESTAR_HEAP SSV_DKG_PORT SIREN_PASSWORD )
+    ERIGON_P2P_PORT_3 LODESTAR_HEAP SSV_DKG_PORT SIREN_PASSWORD LIDO_DV_EXIT_VERSION OBOL_CHARON_CL_ENDPOINTS)
```
yorickdowne, line 1044:
```suggestion
    ERIGON_P2P_PORT_3 LODESTAR_HEAP SSV_DKG_PORT SIREN_PASSWORD LIDO_DV_EXIT_VERSION OBOL_CHARON_CL_ENDPOINTS )
```

ethd
```diff
@@ -1104,7 +1104,8 @@ __env_migrate() {
   for __var in "${__all_vars[@]}"; do
     __value=$(sed -n -e "s/^${__var}=\(.*\)/\1/p" "${__env_file}.source" || true)
     if [ -n "${__value}" ] || [ "${__var}" = "GRAFFITI" ] || [ "${__var}" = "MEV_RELAYS" ] \
-        || [ "${__var}" = "ETH_DOCKER_TAG" ] || [ "${__var}" = "RAPID_SYNC_URL" ]; then
+        || [ "${__var}" = "ETH_DOCKER_TAG" ] || [ "${__var}" = "RAPID_SYNC_URL" ] \
+        || [ "${__var}" = "OBOL_CHARON_CL_ENDPOINTS" ]; then
```
yorickdowne, line 1108: You don't expect this one to be legitimately empty, do you? I think this change can be left as it was

yorickdowne (changes requested):

ethd
cnupy: It is empty when upgrading from a version prior to v18, so the check below never hits. As a result, the value is left at the default value in `default.env` - `http://consensus:5052` after migration. My intention was to migrate it to the value of `OBOL_CL_NODE` - what was previously used in the `lido-obol.yml` `Charon` service.

yorickdowne: Hmm. OK for now; I'll think on whether there's a better way

yorickdowne (approved):

ethd
cnupy: OK

PR #1965, opened 2024-10-28
Switch Erigon users to new repo

yorickdowne:
This new image supports amd64 again, and `latest` tag works

The Erigon team switched from alpine to debian

Unrelated cleanup: Remove unused `wget` from Reth, and Grandine source build

Erigon source build was tested locally, as there is no CI for source builds

### DIFF
--- a/erigon/Dockerfile.binary
+++ b/erigon/Dockerfile.binary
@@ -14,7 +14,16 @@ ARG GID=10002
 
 USER root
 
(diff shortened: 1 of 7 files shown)


pietjepuk2:
LGTM! Only change that stands out is the removal of `--recurse-submodules`, which I presume is no longer needed?

pietjepuk2 (approved):

PR #1968, opened 2024-10-28
MEV_RELAYS can now be multi-line

yorickdowne:
This is a fairly big change to enable one thing: Multi-line values in `.env`. There is one user that has their `MEV_RELAYS` as multi-line, and we kept breaking their `.env` whenever the version of the template changed.

Not touching the config workflow deeply, but that could be next - it'd make a lot of sense to re-use the Lido CSM config flow for all `MEV_RELAYS` purposes, or at least when `MEV_RELAYS` has not been set by the user, and construct multi-line during config.

That'd be a separate PR


yorickdowne:
I am merging this without review, so I can increment the `.env` version for the ssvlabs repo changeover. Without this, I'll break the setup of the one user that has a multi-line value :/

The awk code is GPT-generated. A review is still welcome, we may want to tighten / improve it

PR #1973, opened 2024-10-29
Fix awk regex

yorickdowne:
I only tested on `mawk` and GNU awk complains about `/\"` and rightly so, that should be `/"`, inside `/ ... /` awk will treat `"` as a literal and it doesn't need to be escaped


yorickdowne:
Tested with gawk on debian. Users reported it on Ubuntu. Not waiting for review as this is breaking people's stuff now; but it did work on my local gawk which has the same version as the user's gawk.

PR #1985, opened 2024-11-13
feat: :sparkles: added max peers to Reth (EL)

victorelec14:
Adds the maximum number of peers that can connect to Reth using the `EL_MAX_PEER_COUNT`` variable of the .env

Note: the default value in reth is 100, but I set it to 50 following Geht (EL) configuration, even with the value in my local .env at 40 peers it works fine. If necessary I can change it to 100

docs: https://reth.rs/cli/reth/node.html?highlight=MAX_OUTBOUND_PEERS#reth-node

```txt
--max-outbound-peers <MAX_OUTBOUND_PEERS>
          Maximum number of outbound requests. default: 100
```

image

### DIFF
--- a/reth.yml
+++ b/reth.yml
@@ -84,7 +84,7 @@ services:
       - --authrpc.jwtsecret
       - /var/lib/reth/ee-secret/jwtsecret
       - --max-outbound-peers
-      - ${EL_MAX_PEER_COUNT:-50}
+      - ${EL_MAX_PEER_COUNT:-100}
     labels:
       - metrics.scrape=true
       - metrics.path=/metrics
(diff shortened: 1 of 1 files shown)


reth.yml
```diff
@@ -83,6 +83,8 @@ services:
       - ${EE_PORT:-8551}
       - --authrpc.jwtsecret
       - /var/lib/reth/ee-secret/jwtsecret
+      - --max-outbound-peers
+      - ${EL_MAX_PEER_COUNT:-50}
```
yorickdowne, line 87: Please do set that to the default that the Reth team recommends, that is, 100. Users can always reduce the amount of peers they want by setting this variable in `.env`

yorickdowne (changes requested):

victorelec14 (commented):
its done :)

yorickdowne (approved):
Thanks for this!

PR #1988, opened 2024-11-16
support multiple consensus nodes in nimbus-vc

b0a7:
pass CL_NODE into docker-entrypoint-vc.sh script, parse in script giving ability to pass multiple '--beacon-node=http://blah:5052' to the validator client via comma separated .env CL_NODE in the same way that you can with other validator clients
(refer to https://nimbus.guide/validator-client-options.html)

### DIFF
--- a/default.env
+++ b/default.env
@@ -167,7 +167,6 @@ DISTRIBUTED=
 EL_NODE=http://execution:8551
 # Consensus client address. This could be comma-separated for Lighthouse, Nimbus or Teku VC clients, with failover,
 # or could just be a remote consensus client URL for "validator only" setups.
-# For Nimbus VC client, put the first beacon node here, and use VC_EXTRAS for additional.
 CL_NODE=http://consensus:5052
 # MEV-boost address. This would only be changed for Vouch setups
 MEV_NODE=http://mev-boost:18550
--- a/default.env
+++ b/default.env
@@ -165,7 +165,7 @@ JWT_SECRET=
 DISTRIBUTED=
 # Authenticated execution client endpoint. This default uses the execution node container.
 EL_NODE=http://execution:8551
-# Consensus client address. This could be comma-separated for Lighthouse or Teku VC clients, with failover,
+# Consensus client address. This could be comma-separated for Lighthouse, Nimbus or Teku VC clients, with failover,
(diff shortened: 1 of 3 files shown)


yorickdowne:
This is already supported by `VC_EXTRAS` for Nimbus. `.env` documents this:
"For Nimbus VC client, put the first beacon node here, and use VC_EXTRAS for additional."

Do you see a significant UX improvement by parsing comma-separated for the user, instead?

b0a7:
> This is already supported by `VC_EXTRAS` for Nimbus. `.env` documents this: "For Nimbus VC client, put the first beacon node here, and use VC_EXTRAS for additional."
> 
> Do you see a significant UX improvement by parsing comma-separated for the user, instead?

IMO it is nice to set it up in the same way in the .env for different validator clients.

b0a7:
To elaborate, personally I had used Teku VC before w eth-docker, so I was accustomed to putting the comma separated list of nodes in CL_NODE.  I made this PR because I got an error when I tried to do the same while setting up a new Nimbus VC node.  I didn't notice the adjacent comment in the .env file about Nimbus to be honest.  At the same time though, without seeing the underlying technical reason for why Nimbus VC is setup differently currently (ie the slight difference in how the endpoints are passed to the VCs), naively I don't think the eth-docker user would have reason to expect there to be different ways of doing the same thing between different clients

yorickdowne (changes requested):
This looks good. One style change please: Could you make it `__nodes` and `__node` please. I am trying to switch to a style where `UPPERCASE` is something external to the script, like a var in `.env`, and `__lowercase` are variables local to the script

yorickdowne:
BTW linting: Check out `CONTRIBUTING.md`, we have a pre-commit hook that makes this so much easier to lint on local commit

b0a7:
> This looks good. One style change please: Could you make it `__nodes` and `__node` please. I am trying to switch to a style where `UPPERCASE` is something external to the script, like a var in `.env`, and `__lowercase` is a variable local to the script

done https://github.com/eth-educators/eth-docker/pull/1988/commits/c3aadd5086c1d7ee6d2a184726d7fc8b4a66e5ac
double checking, is test-lodestar the right label and not nimbus?

yorickdowne (approved):
Thank you for this!

PR #1994, opened 2024-11-18
Nimbus EL update

yorickdowne:
Improve Nimbus EL support


pietjepuk2 (approved):

yorickdowne:
Not doing CI yet until we have a Dockerfile.binary. Merging this for now so we have something that compiles and can be used for further testing.

PR #1998, opened 2024-11-26
Switch to ethstaker-deposit-cli

yorickdowne:
Fixes a vuln, supports Python 3.13, supports binary build


yorickdowne:
Merged without review to get the vuln fix in

Future PR would be to clean up deposit-cli.yml some, and think about how to support the additional features of ethstaker-deposit-cli

PR #2002, opened 2024-12-09
Update lido-obol.yml

nameisremus:
Update lido-dv-exit sidecar so that it works with charon v1.0.1 and up

Relevant: https://github.com/ObolNetwork/lido-charon-distributed-validator-node/pull/82


yorickdowne:
unless-stopped is important to be sure! Thanks for this

yorickdowne (approved):
Thanks. Good catch on the restart parameter

Issue #1735, opened 2024-02-06
DDNS for v4 and v6


yorickdowne:
IPv4 only right now since upstream doesn't have ipv4-and-ipv6 yet. That's coming: https://github.com/qdm12/ddns-updater/issues/507

yorickdowne:
If or when Edna-updater gains “ipv4 and ipv6” that can be revisited. In the meantime, duplicating config works.

PR #2012, opened 2024-12-16
DDNS updates v6 address

yorickdowne:
By duplicating the config. Tested on dual-stack and IPV4 only. On v4-only, the v6 probe fails and DDNS does not attempt to update the v6 address, as expected.

Addresses #1735 . Duplicating config is not clean, but appears to work with CloudFlare. This can be changed to "ipv4 and ipv6" if or when ddns-updater gains that functionality.


pietjepuk2:
LGTM

pietjepuk2 (approved):
Also clear from your comment why you duplicated it 👍

PR #2016, opened 2024-12-25
Add validator to nimbus-stats.yml

### LINKED ISSUE #2014: Nimbus beaconcha.in stats don't show validator
Needs to be resolved upstream

https://github.com/gobitfly/eth2-client-metrics-exporter/issues/17

yorickdowne:
Only if Nimbus has the required metrics, and upstream implements that. Would close #2014


yorickdowne:
This isn't likely to happen in this form. Instead of eth2-client-metrics-exporter gaining this functionality, Nimbus would need to gain it.

PR #2022, opened 2025-01-05
fix(xatu): Add `sentry` command

samcm:
Fix for:
```
Error: unknown flag --preset
```

Thanks for the heads up @egk10, and for implementing Xatu @yorickdowne 👑


yorickdowne:
Thanks for finding this!

PR #2026, opened 2025-01-11
Update base-config.yml

Vikt0rPavlik:
This change fixes a critical configuration parameter typo in the Prometheus base configuration. The correct parameter name is scrape_configs_files (with an 's'). This parameter is used to include additional configuration files from the conf.d directory. Without this correction, Prometheus might fail to load additional scrape configurations, potentially missing important metrics collection targets.


yorickdowne:
This is not so. See https://prometheus.io/docs/prometheus/latest/configuration/configuration/. The current syntax is correct.

PR #2027, opened 2025-01-11
Add source build CI every Tuesday

yorickdowne:
For every CL and EL client. Can also be triggered manually at need.

### DIFF
--- a/.github/workflows/build-besu.yml
+++ b/.github/workflows/build-besu.yml
@@ -1,22 +1,16 @@
-name: Test Lodestar and Erigon
+name: Source build Besu
 
 defaults:
   run:
     shell: bash
 
 on:
-  push:
-  pull_request:
-    types: [opened, synchronize, labeled, unlabeled]
-    branches: [main]
+  schedule:
+    - cron: "42 7 * * 2"  # Weekly Tuesday at 7:42 AM UTC
+  workflow_dispatch:
 
 jobs:
-  test-lodestar-erigon:
-    if: |
-      contains(github.event.pull_request.labels.*.name, 'test-lodestar') ||
-      contains(github.event.pull_request.labels.*.name, 'test-erigon') ||
-      contains(github.event.pull_request.labels.*.name, 'test-all') ||
-      github.event_name == 'push'
+  build-besu:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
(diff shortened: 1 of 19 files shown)


.github/workflows/build-besu.yml
```diff
@@ -0,0 +1,57 @@
+name: Source build Besu
+
+on:
+  schedule:
+    - cron: "0 0 * * 0"  # Weekly at 12:00 AM UTC
```
pietjepuk2, line 5: Soft recommendation maybe is to pick a more random time. Like, weekly at 14:12 or something. I'm not sure what the consequences are of 0 0, but I think there'll be more than one triggered workflow starting exactly at midnight UTC 😬

.github/workflows/build-besu.yml
```diff
@@ -0,0 +1,57 @@
...
+          COMPOSE_FILE=teku.yml:besu.yml
+          var=COMPOSE_FILE
+          set_value_in_env
+          FEE_RECIPIENT=0xDccf8451070a86183eE70D330C4c43b686E9CF86
+          var=FEE_RECIPIENT
+          set_value_in_env
+          BESU_DOCKERFILE=Dockerfile.source
+          var=BESU_DOCKERFILE
+          set_value_in_env
+      - name: Build Besu
```
pietjepuk2, line 30: There's a _lot_ of duplicate code here. (the docker buildx part above maybe a bit less, but even that occurs many times). 

If this works, maybe add a commit on top to add an action. So make a `.github/actions/test_source_build/action.yml` with:


```yaml
name: "Test Source Build"
description: "A custom action to test the VC/EL source builds"
inputs:
  command:
    description: "The command to run"
    required: true

runs:
  using: "composite"
  steps:
  - name: Build client
    run: ./ethd update
    shell: bash
  - name: Start client combo
    run: ./ethd up
    shell: bash
  - name: Pause for 30 seconds
    run: sleep 30
    shell: bash
  - name: Test CL
    run: ./.github/check-service.sh consensus
    shell: bash
  - name: Test VC
    run: ./.github/check-service.sh validator
    shell: bash
  - name: Test EL
    run: ./.github/check-service.sh execution
    shell: bash
  - name: Set client combo w/ VC
    run: ${{ inputs.command }}
    shell: bash
  - name: Start client combo
    run: ./ethd up
    shell: bash
  - name: Pause for 30 seconds
    run: sleep 30
    shell: bash
  - name: Test CL
    run: ./.github/check-service.sh consensus
    shell: bash
  - name: Test VC
    run: ./.github/check-service.sh validator
    shell: bash
  - name: Test EL
    run: ./.github/check-service.sh execution
    shell: bash
    ```
    
And using it like 

```yaml
      - name: Run Deploy Routine
        uses: ./.github/actions/test_source_build
        with:
          command: |
            source ./.github/helper.sh
            COMPOSE_FILE=teku-cl-only.yml:teku-vc-only.yml:besu.yml
            var=COMPOSE_FILE
            set_value_in_env
```

pietjepuk2 (changes requested):

.github/workflows/build-besu.yml
yorickdowne: I don't think we need to. This isn't time-sensitive and Github will schedule workflows to fit into our allotment. 

Max 20 concurrent jobs, max 6 hours per job

https://docs.github.com/en/actions/administering-github-actions/usage-limits-billing-and-administration

.github/workflows/build-besu.yml
yorickdowne: Great suggestion, thank you. I had no idea custom actions were a thing.

I DRY'd it a little differently - this way it also applies to the test commands, and is usable for things like Grandine, at least partially.

I could have the ethd update in the action as well if you prefer, that way it always runs even when not source-building.

yorickdowne: Moved ethd update into the action.

Grandine is a special case because it doesn't have a VC. That's OK - once they have one, or if they ever do, this action can be used.

yorickdowne: Crud no that was a mistake. It'd run the source build twice, and that's definitely not what I want. Hang loose.

yorickdowne: Undid that. This is likely as good as it gets right now.

.github/workflows/build-erigon.yml
```diff
@@ -0,0 +1,42 @@
...
+          set_value_in_env
+          FEE_RECIPIENT=0xDccf8451070a86183eE70D330C4c43b686E9CF86
+          var=FEE_RECIPIENT
+          set_value_in_env
+          ERIGON_DOCKERFILE=Dockerfile.source
+          var=ERIGON_DOCKERFILE
+          set_value_in_env
+      - name: Build clients
+        run: ./ethd update
+        shell: bash
```
pietjepuk2, line 32: Lingering from the actions? I don't think setting shell is required in workflows.

pietjepuk2 (changes requested):

.github/workflows/build-besu.yml
yorickdowne: Some more changes around bash defaults.

pietjepuk2 (approved):

PR #2033, opened 2025-01-16
feat: Add ethPandaOps Contributoor

samcm:
This PR adds support for ethPandaOps Contributoor, which is a like-for-like replacement with Xatu Sentry. Contributoor is a more streamlined and slim version of Xatu Sentry - with a reduced footprint and dependancy tree. Contributoor is also much easier to install and maintain for the majority of users, with an `installer` enshrined via https://github.com/ethpandaops/contributoor-installer (although eth-docker doesn't need to make use of this!)

Read more: https://github.com/ethpandaops/contributoor

I've left Xatu Sentry in there for now as some may be using it, but the preference would be to remove it after a little while.


default.env
```diff
@@ -351,4 +359,4 @@ NODE_EXPORTER_IGNORE_MOUNT_REGEX='^/(dev|proc|sys|run|var/lib/docker/.+)($|/)'
 DOCKER_ROOT=/var/lib/docker
 
 # Used by ethd update - please do not adjust
-ENV_VERSION=22
+ENV_VERSION=23
```
yorickdowne, line 362: As we discussed, let me work on migration. That's easier if this version is not incremented here, and instead gets incremented with the migration PR

yorickdowne (changes requested):

default.env
yorickdowne, line 362:
```suggestion
ENV_VERSION=22
```

yorickdowne (changes requested):

yorickdowne (approved):

PR #2034, opened 2025-01-19
rename ERIGON_P2P_PORT_2 -> EL_P2P_PORT_2

b0a7:
Looking through the Reth code I noticed that one of the parameters in use is from the Erigon section of .env.  For clarity and consistency, this PR renames the variable to be more generic, following the EL_P2P naming of the other variable that is shared amongst multiple clients.  IMO it is a bit sloppy to use an Erigon named variable in the Reth code.


default.env
```diff
@@ -119,8 +121,7 @@ EL_RPC_PORT=8545
 EL_WS_PORT=8546
 # Erigon's torrent port. Don't make this 42070, as it will fail
 ERIGON_TORRENT_PORT=42069
-# Erigon's second and third P2P port, used for multiple eth/xx P2P protocols
-ERIGON_P2P_PORT_2=30304
+# Erigon's third P2P port, used for multiple eth/xx P2P protocols
 ERIGON_P2P_PORT_3=30305
```
b0a7, line 125: is the 'used for multiple eth/xx P2P protocols' referring to the usage in Reth?  If so I can remove it here

yorickdowne: Yes that could be removed here

yorickdowne:
Reasonable. This would then also require a change in `ethd` to have the old name under __old_vars and the new under __new_vars, so any custom value a user set would be migrated

default.env
```diff
@@ -119,8 +121,7 @@ EL_RPC_PORT=8545
 EL_WS_PORT=8546
 # Erigon's torrent port. Don't make this 42070, as it will fail
 ERIGON_TORRENT_PORT=42069
-# Erigon's second and third P2P port, used for multiple eth/xx P2P protocols
-ERIGON_P2P_PORT_2=30304
+# Erigon's third P2P port, used for multiple eth/xx P2P protocols
```
yorickdowne, line 124:
```suggestion
# Erigon's third P2P port
```

yorickdowne (changes requested):
Add `ERIGON_P2P_PORT_2` to `__old_vars` and `EL_P2P_PORT_2` to `__new_vars` in `.env`

default.env
yorickdowne, line 1: You also need to increment the version of `default.env` to `24` so that this migration takes place for users.

yorickdowne (changes requested):

b0a7:
> Add `ERIGON_P2P_PORT_2` to `__old_vars` and `EL_P2P_PORT_2` to `__new_vars` in `.env`

fixed https://github.com/eth-educators/eth-docker/pull/2034/commits/f84a76bcffb39e6bb1c78eac335235346d2b8607 (__old_vars/__new_vars is in ethd, not .evn)

yorickdowne (approved):
Looks good, thanks!

PR #2043, opened 2025-01-25
config JWT_SECRET improvements

b0a7:
A couple little tweaks to JWT_SECRET input

a) accept and strip 0x prefix
b) validate only hex chars are in the input


yorickdowne (approved):
Looks good to me

PR #1822, opened 2024-05-05
Remove ext-network.yml

yorickdowne:

PR #2045, opened 2025-01-28
Fix ext-network.yml prep

yorickdowne:
The way I am handling ext-network.yml is not good. What it should be is with a variable in `.env` like everything else

Fixing this tech debt needs to be two-step: One before Pectra, one a month after when everyone has updated and has the code in `ethd` that removes the "don't track" and so the new `ethd` that gets executed on update on the second run (ETHDSECUNDO) can just not copy the .bak in and not untrack, and we are solid.

It's a bit opaque maybe, this was the best I could come up with to resolve it. The flow is:

- Users upgrade pre- or post-Pectra to stay on the chain. A new ethd comes in. For users coming from old Eth Docker nothing happens to ext-network.yml, as the first run is still the old ethd and the second run doesn't find a .bak file
- If they update again and they have this ethd in it, info is extracted and put into .env, a copy is created, and the copy is then copied back during second run so nothing changes for them.

- Users upgrade at some point post-Pectra, at least one month after, and another set of changes comes in
- They will now execute the first run parts - extract info from ext-network and stick it in .env, copy to a bak file, etc
- The second-run part will have changed to just remove the bak file if present, and ext-network will have changed to use the created variable

The part after Pectra is #1822 . That one will need to be updated after this is merged, as otherwise I don't have the ethd reference.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,8 +1,7 @@
 .env*
 docker-compose.yml
-ext-network.yml
-ext-network.yml.original
-ext-network.yml.bak
+*.original
+*.bak
 .eth/*
 .eth_backup*
 !.eth/README.md
@@ -14,6 +13,7 @@ ext-network.yml.bak
 !.eth/ethdo/README.md
 !.eth/ethdo/create-withdrawal-change.sh
 *.swp
+custom.yml
 ssv-config.yaml
 ssv-config.yaml.bak
 blox-ssv-config.yaml
--- a/default.env
+++ b/default.env
@@ -91,6 +91,9 @@ SHARE_IP=
 # Relays to connect charon node
 OBOL_P2P_RELAYS=
 
+# External Docker network if using ext-network.yml
+DOCKER_EXT_NETWORK=rocketpool_net
+
 # P2P ports you will forward to your staking node. Adjust here if you are
 # going to use something other than defaults.
 EL_P2P_PORT=30303
@@ -356,4 +359,4 @@ NODE_EXPORTER_IGNORE_MOUNT_REGEX='^/(dev|proc|sys|run|var/lib/docker/.+)($|/)'
 DOCKER_ROOT=/var/lib/docker
 
 # Used by ethd update - please do not adjust
-ENV_VERSION=25
+ENV_VERSION=26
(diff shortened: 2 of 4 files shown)


yorickdowne:
Copy-pasta of the rationale behind this ... what on earth is Yorick thinking

What I am trying to do is change from a hard coded value in ext-network.yml to a variable
But because I foolishly originally thought I had to change this with sed instead of using a variable - and don’t ask me why I thought that or maybe an earlier compose actually required it - the thing is “not tracked”. And potentially changed on machines that integrate with rocketpool
So I cannot make a change to that file upstream without breaking the users workflow - git pull will say there is a local conflict and be unable to resolve it 
So what I must do:
Undo the assume-unchanged
extract the value and stick it into .env
git restore the file, git pull, so I get the new file without conflict warnings, in case it was changed locally

And that’s where it gets messier 
What code does that? The user has an ethd that doesn’t do that, the code to do that would come in with git pull. Right.
So. I have to have the ethd code in place already that undoes the assume-unchanged, extracts the value, sticks it into .env, restores and pulls , if I want to change ext-network in the repo

But it’s not in place, it comes in with the next git pull, meaning the next ethd update 

ethd for this reason runs its update process twice … once up to git pull and then the rest. This works great for ethd update changes that happen after git pull, but not for ones that come before 

Therefore. I put this code in now but because ext-network is actually unchanged in the repo, an ethd update on an older version won’t break during the git pull.

A future ethd update would then have this code before git pull already in place. So it will successfully remove assume unchanged, restore the file, pull

The current ethd on second execution copies the backup back in and assumes unchanged again, the part after git pull. That’s so a user can update once, get this code, and still update again, and not run into issues

I must be sure all users have updated and have the pre-git-pull code. So I wait until one month after Pectra
At that point I make another change:

ethd code before the git pull no longer special cases ext-network. Ethd code after git pull just removes any bak files should they exist 

The flow then is:
the code brought in as part of Pectra update executes, the portion up to git pull. It can now successfully pull a changed ext-network.yml

This brings in the post-Pectra ethd and executes its second run

Which no longer copies the bak file to ext-network.yml, merely deletes the backup

And then one month post Fusaka I can also clean up the code that looks for .bak and .original of ext-network

It’s a long, long way around to clear this tech debt. And it can only happen around hard forks, because I must be sure that users have upgraded once already to get the “ethd prep code so I can safely pull in a changed ext-network”

pietjepuk2 (approved):

PR #2047, opened 2025-01-29
Fix EL_NODE during RocketPool config

yorickdowne:
This crept in at some point: When configuring for RocketPool, `EL_NODE` is set to empty, as it's not assigned a value. The next `./ethd update` then brings the default in.

This sets a default in ethd.

### DIFF
--- a/ethd
+++ b/ethd
@@ -3928,7 +3928,7 @@ config() {
   __var=COMPOSE_FILE
   __update_value_in_env "${__var}" "${!__var-}" "${__env_file}"
   __var=EL_NODE
-  __update_value_in_env "${__var}" "${!__var-}" "${__env_file}"
+  __update_value_in_env "${__var}" "${!__var:-"http://execution:8551"}" "${__env_file}"
   __var=JWT_SECRET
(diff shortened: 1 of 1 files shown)


ethd
```diff
@@ -3928,7 +3928,7 @@ config() {
   __var=COMPOSE_FILE
   __update_value_in_env "${__var}" "${!__var-}" "${__env_file}"
   __var=EL_NODE
-  __update_value_in_env "${__var}" "${!__var-}" "${__env_file}"
+  __update_value_in_env "${__var}" "${!__var:-"http://execution:8551"}" "${__env_file}"
```
pietjepuk2, line 3931: Am I correct in understanding that for everything but rocketpool, the EL_NODE value is set with `__query_execution_client` during install?

yorickdowne: That is correct

jackvreeken (approved):

ethd
jackvreeken: Approved then!

pietjepuk2 (approved):

PR #2049, opened 2025-01-29
More Nimbus source build fixes, notably Gnosis

yorickdowne:
I had the wrong target for Gnosis VC

`cmake` is from Discord discussions. I don't think current Nimbus build uses it ... if they ever do, it's now there

### DIFF
--- a/nimbus/Dockerfile.source
+++ b/nimbus/Dockerfile.source
@@ -10,10 +10,10 @@ ARG DOCKER_VC_REPO
 ARG BUILD_TARGET
 ARG SRC_REPO
 
-RUN apt-get update && apt-get install -y build-essential bash git-lfs
+RUN apt-get update && apt-get install -y build-essential ca-certificates bash git-lfs cmake
 
 WORKDIR /usr/src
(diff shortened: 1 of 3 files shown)


nimbus/Dockerfile.source
```diff
@@ -10,10 +10,10 @@ ARG DOCKER_VC_REPO
 ARG BUILD_TARGET
 ARG SRC_REPO
 
-RUN apt-get update && apt-get install -y build-essential bash git-lfs
+RUN apt-get update && apt-get install -y build-essential ca-certificates bash git-lfs cmake
```
pietjepuk2, line 13: Why did the source build of Nimbus this week succeed? Are the workflows not checking enough? You say "build fixes" , but the source build was not broken no?

nimbus/Dockerfile.source
```diff
@@ -10,10 +10,10 @@ ARG DOCKER_VC_REPO
...
 
 WORKDIR /usr/src
-RUN bash -c "git clone --recurse-submodules -j8 ${SRC_REPO} nimbus-eth2 && cd nimbus-eth2 && git config advice.detachedHead false && git fetch --all --tags && if [[ ${BUILD_TARGET} =~ pr-.+ ]]; then git fetch origin pull/$(echo ${BUILD_TARGET} | cut -d '-' -f 2)/head:nim-pr; git checkout nim-pr; else git checkout ${BUILD_TARGET}; fi && make USE_SYSTEM_NIM=1 -j$(nproc) update && make -j$(nproc) nimbus_beacon_node nimbus_validator_client"
```
pietjepuk2, line 16: Where is Nim coming from now? Is it built as part of the `make`?

nimbus/Dockerfile.source
yorickdowne: This was not broken. Gnosis was. This follows the Nimbus team recommendation who say they need cmake for some of the dependencies.

lest it break in future, I added cmake

nimbus/Dockerfile.source
yorickdowne: As it was before, this just makes it clearer now that that's what's happening. make update builds Nim

pietjepuk2 (approved):

PR #2051, opened 2025-01-30
Improve sending change message with ethdo

yorickdowne:
- Gives a count of validators found when creating the change json
- Turns on `--debug` so the user can see where the process fails, should it time out: `ethdo` cannot just start over, it needs to start from the first validator that wasn't successful

### DIFF
--- a/.eth/ethdo/create-withdrawal-change.sh
+++ b/.eth/ethdo/create-withdrawal-change.sh
@@ -1,25 +1,29 @@
 #!/usr/bin/env bash
 cd "$(dirname "$(realpath "${BASH_SOURCE[0]}")")"
-set -e
-echo "Copying ethdo to home directory, ~/ethdo"
-mkdir -p ~/ethdo
-cp ethdo ~/ethdo/
-cp ethdo-arm64 ~/ethdo/
-chmod +x ~/ethdo/*
-set +e
 
 __arch=$(uname -m)
-
-if [ "${__arch}" = "aarch64" ]; then
-    __ethdo=~/ethdo/ethdo-arm64
+if [[ "${__arch}" = "aarch64" || "${__arch}" = "arm64" ]]; then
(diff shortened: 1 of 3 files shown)


.eth/ethdo/create-withdrawal-change.sh
```diff
@@ -114,3 +114,9 @@ echo "change-operations.json can be found on your USB drive"
 echo
 echo "Please shut down this machine and continue online, with the created change-operations.json file"
 echo "You can submit it to https://beaconcha.in/tools/broadcast, for example"
+echo
+echo "Attempting to read the number of messages in the change-operations.json"
+if ! type -P jq >/dev/null 2>&1; then
+  sudo apt-get update && sudo apt-get install -y jq
```
pietjepuk2, line 120: But.. aren't we offline?

pietjepuk2 (changes requested):
~~A more general question about this workflow: ethdo needs a file with all information on all validators to work offline no? Where is that code, or is that part of some instructions somewhere?~~ In `.eth`

.eth/ethdo/create-withdrawal-change.sh
yorickdowne: That is indeed a very good point. We are offline. D’uh.

ok. I still think it’s useful to know how many messages were created. Maybe if I download jq and bundle it with ethdo during the offline prep step?

pietjepuk2: Yep, that sounds fine! With the hint of "offline prep step" I managed to find the docs to answer my other question :)

yorickdowne: Change made, jq is now being downloaded. Tested the offline prep and change message flow locally on amd64 and that worked.

pietjepuk2 (approved):

PR #2056, opened 2025-02-06
Split EL and CL archive node

yorickdowne:
Most people who want an archive just want the EL

For CL, it makes sense to also keep blobs, for L2 syncing

Therefore split this into two env vars, and for those CLs that can, keep all blobs

More granular settings for CL can be done with CL_EXTRAS and highly depend on the specific CL


ethd
```diff
@@ -1222,8 +1222,8 @@ __env_migrate() {
     SIREN_DOCKER_TAG SIREN_DOCKER_REPO SSV_DKG_TAG NODE_EXPORTER_IGNORE_MOUNT_REGEX DOCKER_ROOT DEPCLI_DOCKERFILE \
     CONTRIBUTOOR_DOCKER_REPO CONTRIBUTOOR_DOCKER_TAG)
 
-  __old_vars=( XATU_KEY ERIGON_P2P_PORT_2 OBOL_EL_NODE )
-  __new_vars=( CONTRIBUTOOR_USERNAME EL_P2P_PORT_2 EL_RPC_NODE )
+  __old_vars=( XATU_KEY ERIGON_P2P_PORT_2 OBOL_EL_NODE ARCHIVE_NODE )
+  __new_vars=( CONTRIBUTOOR_USERNAME EL_P2P_PORT_2 EL_RPC_NODE EL_ARCHIVE_NODE )
```
pietjepuk2, line 1226: Shouldn't we also add "ARCHIVE_NODE" -> "CL_ARCHIVE_NODE" here? If that's possible

yorickdowne: Not here but elsewhere I can, yes. It's a good question.

CL Archive is incredibly niche, and really only useful if also keeping all blobs - for L2 sync without L2 snapshot or with L2 snapshot older than 18 days.

I'm adding blob archive in this release. My assumption here is that the few users - I know of 1 - who set `ARCHIVE_NODE`, actually want the EL archive.

Those who also want a CL archive could set that in .env or use `CL_EXTRAS` for more granular archive parameters.

A CL that has been synced as archive will, depending on client, remain an archive even without this parameter set.

I am inconveniencing someone either way. I assume EL archive users are more numerous than CL archive users, and so not migrating the CL archive setting (and thus turning on blob retention on the clients that support that) is friendlier overall.

pietjepuk2: Sounds fair enough, and I agree that most users care about EL archiving.

pietjepuk2 (approved):

PR #2058, opened 2025-02-10
Change postgres upgrade image

yorickdowne:
The pats22 image does not appear to be supported any more.

Switch to https://github.com/pgautoupgrade/docker-pgautoupgrade , which upgrades in place, and run that in "one-shot" mode

A good time to merge would be February 14th, right after the release of PostgreSQL 17.3

### DIFF
--- a/ethd
+++ b/ethd
@@ -7,7 +7,7 @@ __sample_service="consensus"
 __docker_exe="docker"
 __compose_exe="docker compose"
 __compose_upgraded=0
-__target_pg=16
+__target_pg=17
 
 
 __dodocker() {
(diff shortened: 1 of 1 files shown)


pietjepuk2:
Looks good! Also looks a lot simpler than the previous approach 👍

yorickdowne:
Great thank you

There is one change I still want to make, which will be a separate PR: Run this in screen, if screen is available. On Debian and Ubuntu, install screen if it's not there.

That would mean that this survives an SSH disconnect, which seems prudent for something that touches the slashing protection DB

PR #2062, opened 2025-02-12
Fix permissions if ~ is 750, e.g Lynis

yorickdowne:
A user ran Lynis hardening, which sets `~` to 750 permissions. This breaks assumptions that SSV and Prometheus make about their bind mounts.

When `ethd` starts, check permissions of the `prometheus` and `ssv-config` directories and fix as needed. We still assume it's owned by the executing user - if that's not the case, a whole lot more would break anyway.


grafana-rootless.yml
```diff
@@ -21,7 +21,8 @@ services:
       - ./prometheus/custom-prom.yml:/etc/prometheus/custom-prom.yml:ro
     environment:
       - CLIENT=${COMPOSE_FILE}
-    command:
+    entrypoint:
```
pietjepuk2, line 24: What is the purpose of these changes? Are they also required to fix the permissions issue?

yorickdowne: They are not required to fix the permissions issue. They do bring the prometheus:local image more to how other images work, where it’s possible to understand what’s being run by looking at the yml, without needing to know what entrypoint is set in the Dockerfile

yorickdowne: Fixed that, you are right.

yorickdowne:
Merging without final review bcs of travel

PR #2063, opened 2025-02-13
ethd update runs in screen if interactive

yorickdowne:
ethd update is long-running, and it's best if it can complete. This is especially true when upgrading the web3signer DB, but holds generally.

Along the same lines, ethd update should only ever have one instance of it running at a time, for any given Eth Docker install


yorickdowne:
Merging without review bcs of travel and PostgreSQL 17.3 update. However, this is worth a discussion - is running in screen for update strictly a good thing, or should we narrowly only do it when migrating PostgreSQL?

PR #2066, opened 2025-02-14
Fix typographical error

PixelPil0t1:
The word "Contributoor" is a typographical error. The correct term is "Contributor," which is the standard English word to describe someone who contributes, especially in contexts like open-source software or collaborative work.


yorickdowne:
Heh. Try again, AI.

PR #2087, opened 2025-03-08
Set Vero stop grace period to 1m

yorickdowne:
From Vero sample compose yml:

    # It's recommended to set the `stop_grace_period` value
    # (or equivalent, e.g. `terminationGracePeriodSeconds` in k8s)
    # to at least 30 seconds.
    # This is because Vero defers the shutdown process if it detects
    # an upcoming validator duty, opting to complete the duty before
    # shutting down.
    # This feature can make the shutdown process take several seconds.

### DIFF
--- a/vero-vc-only.yml
+++ b/vero-vc-only.yml
@@ -23,6 +23,7 @@ services:
     image: vero:local
     pull_policy: never
     user: vero
+    stop_grace_period: 1m
     environment:
       - MEV_BOOST=${MEV_BOOST}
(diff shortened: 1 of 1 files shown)


pietjepuk2:
LGTM, just following recommendations. Interesting that they wait for incoming duties; I'm not sure if they mean all duties (incl attestations), or just proposal/sync committee

pietjepuk2 (approved):

PR #2088, opened 2025-03-08
Calling docker-entrypoint-vc as root calls itself after chown

yorickdowne:
The entrypoint scripts have a feature where they'll chown the datadir and call themselves with exec gosu non-root-user - this is intended as a way for users to relatively easily fix permissions issues, if they have to.

Clearly it's never been used by anyone: `docker-entrypoint-vc.sh` would have tried to call `docker-entrypoint.sh`, which is the entrypoint for the consensus client.

### DIFF
--- a/lighthouse/docker-entrypoint-vc.sh
+++ b/lighthouse/docker-entrypoint-vc.sh
@@ -3,7 +3,7 @@ set -Eeuo pipefail
 
 if [ "$(id -u)" = '0' ]; then
   chown -R lhvalidator:lhvalidator /var/lib/lighthouse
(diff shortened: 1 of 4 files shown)


pietjepuk2:
I guess this showed up with Vero (or inspired by it)? Apparently not many "vc-only" users

pietjepuk2 (approved):

yorickdowne:
Yeah noticed it when working on Vero

there are some VC only users, notably reverse hybrid with RocketPool. But they don’t run as root -  no one does. This code path exists so users could temporarily change to running as root to resolve permissions issues

PR #2105, opened 2025-03-18
fix: double added --Prunning.Mode for Nethermind

madlabman:
"Option '--Pruning.Mode' expects a single argument but 2 were provided" error if there's ARCHIVE_NODE=true and a custom network.


yorickdowne (approved):
Thank you. Hoodi is a blessing - the custom code path gets some testing :rofl:

PR #2109, opened 2025-03-19
Remove CL/EL match nag

yorickdowne:
This was useful around merge, but is no longer


yorickdowne:
I am deliberately not incrementing the .env version here. Removing `DISTRIBUTED` from user `.env` isn't urgent and can happen next time there's a change to `default.env` that must be brought across

pietjepuk2 (approved):

PR #2110, opened 2025-03-22
Introduce alias for web3signer

yorickdowne:
Consider a deployment where holesky and hoodi are on one machine, both use web3signer and mev-boost, and both are connected to the same docker bridge network for central metrics collection.

This would currently break:
- web3signer:9000 is hard-coded in multiple locations
- mev-boost doesn't have an alias

Fix it by:
- Introducing an alias for mev-boost, which can then be referenced in `MEV_NODE`
- Introduce `W3S_NODE` and use it throughout. This means changes to almost every VC, and changes to the keymanager script

### DIFF
--- a/commit-boost-pbs.yml
+++ b/commit-boost-pbs.yml
@@ -14,6 +14,10 @@ services:
       CB_METRICS_PORT: 10000
     volumes:
     - ./commit-boost/cb-config.toml:/cb-config.toml:ro
+    networks:
+      default:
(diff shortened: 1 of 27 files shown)


pietjepuk2 (approved):

yorickdowne:
Tested ethd keys list, import and delete all, those all worked

Will merge this when the MINIMAL_NODE split gets merged, they both use version 30

PR #2118, opened 2025-03-24
Support pre-merge history expiry

yorickdowne:
For the clients that have it. Notably, Besu, Reth and Erigon do not, yet. There'll be a separate PR for how Erigon handling changes.


yorickdowne:
Unsurprisingly doesn't actually work yet: `chain not far enough past merge block, need 1540409 more blocks`

Let me put this in draft until Geth is further along and this can be fully tested

PR #2122, opened 2025-03-26
ethd update reads variables from default.env

yorickdowne:
Get rid of `__all_vars` and `__target_vars` entirely

Detect build targets by names ending in `_TAG`, `_REPO`, `_TARGET` or `_DOCKERFILE`

Needs #2121 so Lido image tag variables follow that convention

Also likely conflicts with #2121 so should be rebased after that's been merged

### DIFF
--- a/ethd
+++ b/ethd
@@ -110,7 +110,7 @@ __handle_docker() {
     __update_value_in_env "${__var}" "${!__var}" "${__env_file}"
 
     __var=NODE_EXPORTER_IGNORE_MOUNT_REGEX
-    __value=$(__get_value_from_env "${__var}" "${__env_file}")
+    __get_value_from_env "${__var}" "${__env_file}" "__value"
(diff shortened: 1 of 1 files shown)


yorickdowne:
Why am I assigning to a variable directly in `__get_value_from_env`? The reason is that I want to know whether the value was `__found` in whatever file I am parsing through. If I do like before with `__variable=$(__get_value_from_env SOME_VAR .env)` then I am in a subshell, and while I can return the value, whatever I assign to `__found` stays in the subshell and isn't accessible. Womp womp.

So, to avoid that: No subshell, assign directly. The drawback is I now need to exempt the local variables from legal variable names to assign to, as otherwise the assignment isn't readable outside the function. That's a minor annoyance though, and as this is just a helper function for ethd code, seems acceptable.

pietjepuk2 (approved):
Great cleanup!

PR #2135, opened 2025-04-01
Migration failed with empty values

yorickdowne:
During testing, noticed that empty values don't come across. That needs to be fixed.


yorickdowne:
This is breaking things now, merging without review

PR #2139, opened 2025-04-05
Add Vero support to `./ethd version`

eth2353:
Vero supports the `--version` flag in the v1 release candidates so adding it here.


yorickdowne (approved):
Thanks!

PR #2145, opened 2025-04-12
Better ncdu message

yorickdowne:
A user remarked that she found it confusing to be told to use `ncdu`, but `ncdu` wasn't installed.

Do two things:
- `./ethd install` installs ncdu - it's 100KB and useful
- The space message checks whether `ncdu` is installed, and if not, tells the user how to

### DIFF
--- a/ethd
+++ b/ethd
@@ -332,7 +332,7 @@ install() {
   fi
   if [[ "$__distro" = "ubuntu" ]]; then
     ${__auto_sudo} apt-get update
-    ${__auto_sudo} apt-get install -y ca-certificates curl gnupg whiptail chrony pkg-config screen
+    ${__auto_sudo} apt-get install -y ca-certificates curl gnupg whiptail chrony pkg-config screen ncdu
     echo
     echo
     if [ -z "$(command -v docker)" ]; then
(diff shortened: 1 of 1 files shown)


ethd
```diff
@@ -486,7 +486,18 @@ __display_docker_volumes() {
...
   fi
-  echo "If there is some mystery space being taken up, try \"sudo ncdu /\"."
+  if command -v ncdu >/dev/null 2>&1; then
+    echo "If there is some mystery space being taken up, try \"sudo ncdu /\"."
+  else
+    echo "If there is some mystery space being taken up, install ncdu, then try \"sudo ncdu /\"."
+    if [[ "$OSTYPE" == "darwin"* ]]; then
+      echo "To install ncdu, run \"brew install ncdu\"."
+    elif [[ "$__distro" = "ubuntu" || "$__distro" =~ "debian" ]]; then
+      echo "To install ncdu, run \"sudo apt update && sudo apt install ncdu\"."
```
pietjepuk2, line 496: I assume this is for existing users that won't run "install" again right?

pietjepuk2 (approved):

ethd
yorickdowne: Correct, or for users that never ran install and preferred to set Docker up themselves

PR #2146, opened 2025-04-12
Consistently use = in test statements, not ==

yorickdowne:
`ethd` mostly does `if [ "$a" = "b" ]` but in some locations does `if [ "$a" == "b" ]`

Changing it to always be the former doesn't change function, just makes style consistent


pietjepuk2 (approved):
I like consistency 👍

PR #2165, opened 2025-04-25
Use Teku for Ephemery

yorickdowne:
Lodestar was throwing errors, reported by a user on Discord

As per https://github.com/ephemery-testnet/ephemery-resources/blob/master/client-implementations.md we'd need a special build for Lodestar

Teku and Besu are ready with main images, tested locally

Users who want an experimental client are advanced and can source-build via `.env`, `./ethd config` should offer a seamless experience


pietjepuk2 (approved):

yorickdowne:
Changed it to put Lodestar back in: Dev explained, Lodestar works, but on Ephemery reset requires `./ethd resync-consensus`, whereas Teku does not.

That's still good enough to have it in the menu

PR #2167, opened 2025-04-26
Automatically change old Besu tags

yorickdowne:
These tags were last updated one year ago, the current tag is `latest`. A user was running with `latest-openjdk-latest` and they got "Mantic" 404s :/


yorickdowne:
Merging without review as this is likely to hit other users

PR #2183, opened 2025-05-12
[OBOL] Fix metrics exporter CL url

vgorkavenko:
Fixing the issue where the exporter was collecting metrics from `charon`, not from CL node


yorickdowne:
consensus is always the consensus layer client service, not charon.

Do you mean this changes things so it does collect from charon? CL_NODE is used by the VC, and so this change would have it collect data from charon instead of from the consensus layer client - if I understand the Lido X Obol setup and ETH docker right

yorickdowne (approved):
I don't think this'll break anything - it already assumes `execution` to be local, after all

PR #2184, opened 2025-05-13
Support Besu pre-merge history expiry

yorickdowne:
A followup PR will be needed when Besu changes its flags. `--Xsnapsync-synchronizer-pre-merge-headers-only-enabled=true` may become `--snapsync-synchronizer-pre-merge-headers-only-enabled=true` and be default.

For now, if `EL_MINIMAL_NODE=true`, sync with `--Xsnapsync-synchronizer-pre-merge-headers-only-enabled=true`.

`./ethd prune-history` runs `storage prune-pre-merge-blocks` then on restart enables RocksDB garbage collection with `--Xhistory-expiry-prune`. This should free space within 2-6 hours. To be on the safe side, it's kept active for 48 hours, and if Besu restarts after that, the flag is removed.

Docs: https://github.com/hyperledger/besu/releases/tag/25.6.0-RC1

Because Besu may need additional space, I want to offer this to users before they run out entirely. I've added a __min_ver and __extra_msg to inform users of the version they need, and any additional requirements.

### DIFF
--- a/besu/docker-entrypoint.sh
+++ b/besu/docker-entrypoint.sh
@@ -53,6 +53,23 @@ fi
 if [ "${ARCHIVE_NODE}" = "true" ]; then
   echo "Besu archive node without pruning"
   __prune="--data-storage-format=FOREST --sync-mode=FULL"
+elif [ "${MINIMAL_NODE}" = "true" ]; then
+  echo "Besu minimal node with pre-merge history expiry"
+  __prune="--Xsnapsync-synchronizer-pre-merge-headers-only-enabled=true"
+  __timestamp_file="/var/lib/besu/prune-history-timestamp.txt"
+  if [ -f "${__timestamp_file}" ]; then
+    __saved_ts=$(<"${__timestamp_file}")
+    __current_ts=$(date +%s)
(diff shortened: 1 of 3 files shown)


besu/docker-entrypoint.sh
```diff
@@ -53,6 +53,23 @@ fi
...
+  if [ -f "${__timestamp_file}" ]; then
+    __saved_ts=$(<"${__timestamp_file}")
+    __current_ts=$(date +%s)
+    __diff=$((__current_ts - __saved_ts))
+
+    if (( __diff >= 172800 )); then  # 48 * 60 * 60 - 48 hours have passed
+      rm -f "${__timestamp_file}"
+    else
+      echo "Enabling RocksDB garbage collection after history prune. You should see Besu DB space usage go down."
+      echo "This may take 6-12 hours. Eth Docker will keep RocksDB garbage collection on for 48 hours."
```
pietjepuk2, line 69: Shouldn't we tell users that they should restart execution after 2 days? Otherwise besu will just keep running with this flag indefinitely.

pietjepuk2 (approved):

PR #2188, opened 2025-05-14
Remove `CF_ZONE_ID` migration, to fix stdin stealing

yorickdowne:
This broke when changing `__env_migrate` to read all of `default.env` from stdin ... if the `CF_ZONE_ID` migration gets to `curl` in a sub-shell, it steals stdin and the loop terminates early.

Since this migration is scheduled to be removed anyway, I may as well do so now. I could fix it with `< /dev/null`, but simple removal seems easier here.

For `__enable_v6`, I am doing both - fix the stdin stealing *and* also remove it from the read loop

### DIFF
--- a/ethd
+++ b/ethd
@@ -1011,47 +1011,6 @@ __upgrade_postgres() {
 }
 
 
-__lookup_cf_zone() { # Migrates traefik-cf setup to use Zone ID
-  __get_value_from_env "COMPOSE_FILE" "${__env_file}.source" "__compose_ymls"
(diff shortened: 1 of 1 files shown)


yorickdowne:
This is breaking a user now @pietjepuk2 , so another merge without review. Apologies. Will :100: entertain any comments and suggestions post-facto

yorickdowne:
The test failure is unrelated, some issue with the ethstaker-deposit-cli repo. That'll be tackled separately.

PR #2193, opened 2025-05-19
install can apply optimizations

yorickdowne:
**What I did**

Refactored `ethd install` broadly, and added the ability to apply optimizations.

As part of this, `__nag_os_version` was also changed. I was checking the version in multiple places and it was becoming unmaintainable. This function now both nags the user and sets a global, so that code that requires a minimum version can return without making changes.

This new install will need to be tested on Debian and Ubuntu before we declare it OK

Future PRs could add hand-holding for unattended-upgrades, msmtp-mta and email notifications, and ssh public key auth, maybe even "placing ufw in front of Docker" if a public IP is detected.

image

### DIFF
--- a/ethd
+++ b/ethd
@@ -5,9 +5,21 @@ __project_name="Eth Docker"
 __app_name="Ethereum node"
 __sample_service="consensus"
 __docker_exe="docker"
+__old_docker=0
 __compose_exe="docker compose"
+__old_compose=0
 __compose_upgraded=0
 __target_pg=17
+__distro=""
+__os_major_version=""
+__eol_os=0
+__min_ubuntu=22
+__suggest_ubuntu="22.04 or 24.04."
+__upgrade_ubuntu="24.04: https://gist.github.com/yorickdowne/94f1e5538007f4c9d3da7b22b0dc28a4"
+__min_debian=11
+__suggest_debian="12 or 11."
+__upgrade_debian="12: https://gist.github.com/yorickdowne/ec9e2c6f4f8a2ee93193469d285cd54c"
+__reboot_required=0
 
 
 __dodocker() {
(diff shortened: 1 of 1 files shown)


ethd
```diff
@@ -326,122 +337,257 @@ __install_bash_completions() {
   else
     if [ ! -f "$(pkg-config --variable=completionsdir bash-completion)/ethd" ]; then
       echo "Installing bash completions for ethd"
-      ${__auto_sudo} ln -s "$(dirname "$(realpath "${BASH_SOURCE[0]}")")/bash-completion" \
+      ${__auto_sudo} ln -sf "$(dirname "$(realpath "${BASH_SOURCE[0]}")")/bash-completion" \
```
pietjepuk2, line 340: Nice! Exactly an issue I ran into myself before

ethd
```diff
@@ -326,122 +337,257 @@ __install_bash_completions() {
...
+
+install() {
+  if ! [[ "${__distro}" =~ (ubuntu|debian) ]]; then
+    echo "${__project_name} does not know how to install Docker on ${__distro}"
+    return 0
   fi
 
-  if ! [[ "$__distro" = "ubuntu" ]] || [[ "$__distro" =~ "debian" ]]; then
+  __nag_os_version
+  if [ "${__eol_os}" -eq 1 ]; then
```
pietjepuk2, line 543: Not sure if intended, but this _also_ means that, even if the user installs docker themselves, they cannot leverage the "apply optimizations" , and 'shell completion" stuff that `install` also does.

ethd
```diff
@@ -326,122 +337,257 @@ __install_bash_completions() {
...
+    echo "${__project_name} requires an updated ${__distro} to run install"
+    echo "Please install Docker manually, if you do not wish to upgrade your OS"
```
pietjepuk2, line 545: To me, at first read, I interpreted this as "install docker yourself, and then you can run the 'install' command again".

yorickdowne: I can change this to be clearer. The intent is to not run anything install does, as there’s no guarantee any of it will work

pietjepuk2 (approved):

PR #2194, opened 2025-05-19
updated link to prysm repo in `prysm-vc-only.yml`

gap-editor:


yorickdowne:
Sure, updating the defaults makes sense. But in that case, it makes sense broadly, not just in this one file

PR #2197, opened 2025-05-24
chore(env): add pruning warning and EIP-4444 data retention info

victorelec14:
**What I did**

Added a clear warning and descriptive comments in the .env file regarding the use of EL_MINIMAL_NODE and its implications under EIP-4444.

Clarified that enabling minimal mode limits historical data to 1 year and included pruning duration (2–10 hours), disk space requirements (≥50 GB), and validator risks if no fallback is configured.

### DIFF
--- a/default.env
+++ b/default.env
@@ -182,10 +182,15 @@ CL_ARCHIVE_NODE=false
 # Set this to true to sync a minimal CL node - in use with Grandine and Teku, for example
 # CL_ARCHIVE_NODE must be false for this to take effect
 CL_MINIMAL_NODE=true
-# Set this to true to sync a node that does not carry all historical data, see EIP-4444
+# Set this to true to sync a node that does not carry more than 1 year of historical data, see EIP-4444
 # Only meaningful for mainnet and sepolia, may fail on hoodi
 # Erigon also has an \"aggressive\" mode that prunes even more
 # EL_ARCHIVE_NODE must be false for this to take effect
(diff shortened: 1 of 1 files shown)


victorelec14:
@yorickdowne about keeping a 1 year history, I read that in the EIP-444, please see this link

https://eips.ethereum.org/EIPS/eip-4444#why-a-year

```txt
 Why a year?
This proposal sets HISTORY_PRUNE_EPOCHS to 82125 epochs (one earth year). This constant is large enough to provide sufficient room for the Weak Subjectivity Period to grow, and it’s also small enough so as to not occupy too much disk space.
```

About the 50GB minimum for pruning, I got it from RocketPool and from comments from people who have had problems when they ran out of space and needed to resynchronize the chain.

https://docs.rocketpool.net/guides/node/pruning
```txt
With that in mind, the following two conditions are required to successfully prune Geth:

A working fallback node configured
At least 50 GB of free space remaining on your SSD
```

It is also important to inform that the prunning is done (in the link above), no attestation will be done until the process is finished, this may take 2-10 hours depending on the hardware. I think this is a long time that node operators should be aware of.
```txt
WARNING

If you don't have a fallback node configured, your node will stop validating during the pruning process. It will miss all attestations and block proposals until it's finished and has resynced with the network. You will leak ETH due to missed validations during this time!
```


Maybe you could reconsider to apply this PR or where could I put this information ? :)

Thank you!

PR #2252, opened 2025-07-30
[LIDO] feat: add Obol remote Loki url configuration

vgorkavenko:
**What I did**

Added a separate promtail instance to Obol configuration to send logs from main containers (CL, EL, Validator, etc.) with specific labels to a remote Loki URL.

The implementation was inspired by https://github.com/ObolNetwork/lido-charon-distributed-validator-node

I didn't find a way how to get client names to have it in labels :(

### DIFF
--- a/lido-obol.yml
+++ b/lido-obol.yml
@@ -24,7 +24,7 @@ services:
       - CHARON_MONITORING_ADDRESS=0.0.0.0:3620
       - CHARON_BUILDER_API=${BUILDER_API_ENABLED:-true}
       - CHARON_FEATURE_SET_ENABLE=eager_double_linear,consensus_participate
-      - CHARON_LOKI_ADDRESSES=${CHARON_LOKI_ADDRESSES:-http://loki:3100/loki/api/v1/push}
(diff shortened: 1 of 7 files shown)


lido-obol.yml
```diff
@@ -24,7 +24,7 @@ services:
       - CHARON_MONITORING_ADDRESS=0.0.0.0:3620
       - CHARON_BUILDER_API=${BUILDER_API_ENABLED:-true}
       - CHARON_FEATURE_SET_ENABLE=eager_double_linear,consensus_participate
-      - CHARON_LOKI_ADDRESSES=${CHARON_LOKI_ADDRESSES:-http://loki:3100/loki/api/v1/push}
+      - CHARON_LOKI_ADDRESSES=http://loki:3100/loki/api/v1/push${OBOL_CHARON_REMOTE_LOKI_ADDRESSES:+,${OBOL_CHARON_REMOTE_LOKI_ADDRESSES}}
```
yorickdowne, line 27: Was this a mistake? I don't understand what the intent is

yorickdowne (changes requested):

lido-obol.yml
vgorkavenko: This was a copy-paste from the original repository. Since logs are now written to local Loki, I think it would be wrong if when adding `OBOL_CHARON_REMOTE_LOKI_ADDRESSES ` they stop being in local one

yorickdowne (approved):

PR #2262, opened 2025-08-26
Add MEV-boost min-bid configuration to the config wizard

shayanb:
Added the proper window and checks to the config wizard so users can input the min-bid for their MEV-Boost setup.

Screenshot 2025-08-26 at 4 51 12 PM


yorickdowne:
I am inclined not to merge this. There are a lot of knobs in default.env that aren't exposed through the config wizard, and this knob is as niche as they come.

The default behavior by VCs is to have a builder boost factor of 90, which takes care of tx inclusions. Min bid is an option, but not one that is very crucial.

shayanb:
makes sense not to over expose the config wizards to all the nuances. I'm closing this PR.

PR #2265, opened 2025-08-30
Improve AWS traefik

yorickdowne:
**What I did**

Bring in the changes from https://github.com/CryptoManufaktur-io/central-proxy-docker

`aws-traefik.yml` can now update A record as well as CNAMEs, and no longer requires a custom traefik image: But does use a custom DNS updater script. I don't think it updates AAAA record, but that could be added.

Bump default Traefik tag to 3.5, now that 3.5.1 is out


yorickdowne:
Tests
- Run with debug logs
- Privilege drop worked
  `11000      15673  0.2  0.3  52996 48056 ?        Ss   09:18   0:00 python updater.py`
- With AWS CLI
  - v4-only: A created, AAAA says none found
  - v4 and v6: A and AAAA created
  - CNAMEs created
  - traefik can do DNS cert verification: TXT records created and removed, ACME provision starts
- Without AWS CLI, using the key directly
  - v4-only: A created, AAAA says none found
  - v4 and v6: A and AAAA created
  - CNAMEs created
  - traefik can do DNS cert verification: TXT records created and removed, ACME provision starts

PR #2267, opened 2025-08-31
Support pre-confs

yorickdowne:
**What I did**

~~Add support for cb-signer. For better UX, a Python init script does some sanity checking of the config and sets the variables correctly for different clients.~~

After reflection, a better target is support for certain pre-confirmations, and use web3signer as the signing module. This could be pre-confirmations for based rollups, or projects like ethgas.

As of July 2026 support for based rollups remains unclear to me. It is also the area I see most deserving of support.

image

### DIFF
--- a/commit-boost-pbs.yml
+++ b/commit-boost-pbs.yml
@@ -11,9 +11,8 @@ services:
     image: ${CB_PBS_DOCKER_REPO:-ghcr.io/commit-boost/pbs}:${CB_PBS_DOCKER_TAG:-latest}
     environment:
       CB_CONFIG: /cb-config.toml
-      CB_METRICS_PORT: 10000
     volumes:
-    - ./commit-boost/cb-config.toml:/cb-config.toml:ro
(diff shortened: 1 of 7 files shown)


yorickdowne:
This would be for eth-gas. Have to confirm that eth-gas uses standard `commit-boost signer` now, and how to configure https://github.com/ethgas-developer/ethgas-preconf-commit-boost-module

They use a relay now. Rather than do this twice, look at this idea post-Glamsterdam

It might be best to require web3signer when using eth-gas, ignore cb-signer entirely. And then ethgas would need to have a way to work with web3signer

From the CB docs, the cb_signer doesn't run when using web3signer :
"Web3Signer implements the same API as Commit-Boost, so there's no need to set up a Signer module. The parameters needed for the remote signer are:
```
[signer.remote]
url = "https://remote.signer.url
```
"

yorickdowne:
ePBS notes: https://hackmd.io/@thUuBZa2ScW2i3kFbsrOFA/S1XTQgR_Ge

PR #2268, opened 2025-09-05
Builds use Debian trixie

yorickdowne:
**What I did**

Switch to Debian trixie and golang 1.25 for both binary and source builds

Fix binary build issues with Nimbus EL

Besu can't build from 25.8.0 tag, switching default tag to `main`

No changes to existing `.env`, which is deliberate


yorickdowne:
Caplin test failing is unrelated, 14s vs 12s spec, that's likely a Caplin bug

PR #2292, opened 2025-09-24
FIxes issue #2291 - need to use sudo to get disk usage on ancien…

### LINKED ISSUE #2291: Need to use sudo to get disk usage on ancient_dir for space()
Fixes bug in getting space used on ancient_dir.

To reproduce:

1. Move Geth ancient directory from current docker volume (eg on nvme) to a different mount point (eg on SATA) like /mnt/geth/chaindata/ancient/
2. In .env file for ethd, set value for path to ancient_dir, such as: ANCIENT_DIR=/mnt/geth/chaindata/ancient/
3. ethd space now produces an error: du: cannot access '/mnt/geth/chaindata/ancient/': Permission denied
4. To fix: change ethd space() function line 716 to:     ${__auto_sudo} du -sh "${ANCIENT_DIR}"

ckeene:
ethd space produces an error if ancient_dir flag is used for Geth

**What I did**
Changed ethd space() function line 716 to: ${__auto_sudo} du -sh "${ANCIENT_DIR}"

**Related issue**
Fixes #2291

### DIFF
--- a/ethd
+++ b/ethd
@@ -713,7 +713,7 @@ space() {
   __get_value_from_env "${__var}" "${__env_file}" "ANCIENT_DIR"
   if [ -n "${ANCIENT_DIR}" ]; then
     echo
-    du -sh "${ANCIENT_DIR}"
+    ${__auto_sudo} du -sh "${ANCIENT_DIR}"
   fi
 }
 


yorickdowne (approved):
Thanks for the report, and the fix!

PR #2328, opened 2025-10-25
feat(prometheus): make TSDB retention time configurable via PROMETHEUS_RETENTION_TIME

korentzari:
- Add PROMETHEUS_RETENTION_TIME to default.env (default: 40d)
- Use it in grafana.yml: --storage.tsdb.retention.time=${PROMETHEUS_RETENTION_TIME:-40d}
- Backward compatible thanks to the compose default (:-40d)
- Users can override retention (e.g., 15d, 12h) without editing compose files


default.env
```diff
@@ -126,6 +126,8 @@ GRAFANA_PORT=3000
 SIREN_PORT=2443
 # Prometheus port used when exposing directly on host; used for federation
 PROMETHEUS_PORT=9090
+# Prometheus TSDB retention time (36h, 40d, 52w, 1y) – how long metrics are kept.
+PROMETHEUS_RETENTION_TIME=40d
```
yorickdowne: Please change the `ENV_VERSION` at the end so this comes in with `./ethd update`

grafana.yml
```diff
@@ -22,7 +22,7 @@ services:
       - ./prometheus/custom-prom.yml:/etc/prometheus/custom-prom.yml:ro
     command:
       - --storage.tsdb.path=/prometheus
-      - --storage.tsdb.retention.time=40d
+      - --storage.tsdb.retention.time=${PROMETHEUS_RETENTION_TIME:-40d}
```
yorickdowne: Please also make this change in `grafana-rootless.yml`

korentzari: I have update grafana-cloud.yml and grafana-rootless.yml

yorickdowne (changes requested):

PR #2335, opened 2025-11-01
Do not reset PG_DOCKER_TAG

yorickdowne:
**What I did**

The default changed to `17-trixie` from `17-bookworm`, and changing that during `./ethd update --refresh-targets` produces collation errors. Worse, when the default changes to `18-trixie`, it'd break entirely, as the DB would not have been migrated.

When upgrading the DB, use `pgautoupgrade:17-trixie` specifically, instead of `17-debian` and then setting the docker tag to `17-trixie`. The Debian version has to match exactly.

### DIFF
--- a/ethd
+++ b/ethd
@@ -1024,23 +1024,23 @@ __upgrade_postgres() {
   echo "In failure case, do not start Web3signer again, instead seek help on Ethstaker Discord."
   echo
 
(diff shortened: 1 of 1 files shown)


yorickdowne:
Merging without review: This is impacting users now

PR #2345, opened 2025-11-05
Erigon WS port is now separate

yorickdowne:
**What I did**

Erigon used to set the WS port the same as the RPC port. It is now a separate port, like on every other client. Configure the port in `erigon.yml` and remove the warnings from `./ethd config`


yorickdowne:
This might have changed in January 2024 with https://github.com/erigontech/erigon/releases/tag/v2.56.0: "Enable superfluous ws.port flag to fix some Hive RPC tests"

PR #2357, opened 2025-11-07
Teku's build factor is on the CL only

yorickdowne:
**What I did**

Remove build factor from the Teku VC, it's not supported there


yorickdowne:
Will merge without review: This is breaking users now

PR #2359, opened 2025-11-07
CI more .env variations

yorickdowne:
**What I did**

Add tests for `IPV6=true`, `IPV6=false`, and for different node types: `EL_MINIMAL_NODE=false` and `CHECKPOINT_SYNC_URL=`

Not every client supports archive and that's very niche, so not adding that to CI right now


yorickdowne:
Won't-do, this has been superseded by 2360

PR #2364, opened 2025-11-10
Add macOS runner for ethd

yorickdowne:
**What I did**

Add a macOS CI runner for ethd

Fold the update test into `test-ethd`: macOS test takes 35 minutes. It's worthwhile; we don't want to always run it. In addition, we can't have "missed variables" any more since the `__env_migrate` refactor

Don't use `flock` on macOS. It's not installed by default and even with `util-linux` won't be in the `$PATH` by default


yorickdowne:
Lima 2.x support is being worked on: https://github.com/douglascamata/setup-docker-macos-action/issues/53

yorickdowne:
Worked fine for test-update and now fails on test-ethd ... more work to figure out where Docker is unhappy on macOS, now

PR #2371, opened 2025-11-12
Change variables to use {}

yorickdowne:
**What I did**

Change style to use {} around variables, declare local, and use __ for global variables


yorickdowne:
Extensively tested `ethd keys`, all set/get/delete , and full import, as well as exit signing and withdrawal change prep. Tested `./ethd keys delete puybkey` and `./ethd keys delete all`.

Tested deposit-cli-new

Tested ethd config, space, resync-consensus, resync-execution, terminate, install, update. Did not test prune-history, will do that after this is merged.

PR #2388, opened 2025-11-25
Update lido-obol.yml and default.env

nameisremus:
This PR updates the `lido-obol.yml` and `default.env` files to reflect Lido's support of execution layer triggerable withdrawals.

It also updates `lido-obol.yml` to reflect the fact that the `eager_double_linear` and `consensus_participate` features are now default, allowing users to set new, optional features through an `.env` flag.


yorickdowne (changes requested):
Bump default.env version please

PR #2414, opened 2025-12-31
env migration handles multiline variables with single quotes

yorickdowne:
**What I did**

Changed the awk code to handle multi-line single-quote entries, like

```
STATIC_PEERS='[
  "enode:something",
  "enode:thatthing"
]
'
```

This pattern is not in use with Eth Docker, but valid. This is adapted from Linea Docker, where the pattern is in use

Main concern: Touching awk code is tricky. I didn't write it to begin with, AI did. And I didn't write this adjustment, AI did.

### DIFF
--- a/ethd
+++ b/ethd
@@ -1409,13 +1409,13 @@ __get_value_from_env() {
     BEGIN { __found = 0; __value = "" }
 
     # Skip empty lines and comments
-    /^#|^\s*$/ {
+    /^[[:space:]]*($|#)/ {
       next
     }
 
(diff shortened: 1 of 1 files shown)


yorickdowne:
I tested this on an `.env` with:

- Empty var
- Single-line no quotes
- Single-line single-quote
- Single-line double-quote
- Multi-line single-quote
- Multi-line double-quote
- Comments

And everything came over fine, with just the `ENV_VERSION` being changed

pietjepuk2 (approved):

Issue #2373, opened 2025-11-13
Feat: Add Grandine-in-Nethermind support


yorickdowne:
Analogous to Caplin

" we have Grandine integrated into Nethermind. This allows to achieve the top performance because there is no overhead of Engine API HTTP communication. It’s basically Nethermind binary with embedded Grandine. https://hub.docker.com/repository/docker/sirsedev/test-repo/tags/unstable-542a084-nethermind-1.35.2/sha256:77fcbf04829cf9935a0d3a021cc9266d42c56d3e8f1965d0df44656a31fdbb15 "

sauliusgrigaitis:
Just wanted to note that in this setup Nethermind's and Grandine's data dirs stay separate (the same way as Grandine would not be embedded into Nethermind). So it should be very easy to keep the data directories structure in `eth-docker` in the way that allows users to switch between separate Grandine/Nethermind and embeded Grandine into Nethermind without resync.

PR #2416, opened 2026-01-04
Initial support for Grandine plugin in Nethermind

yorickdowne:
**What I did**

Support Grandine plugin, with Nethermind binary build only, and for users that explicitly seek it. This requires adjustments in `.env`:

```
COMPOSE_FILE=grandine-plugin.yml:nethermind.yml
# Could also be grandine-plugin-allin1.yml; plus add whatever else the user runs such as commit-boost-pbs, grafana, &c
CL_NODE=http://execution:5052
NM_DOCKER_REPO=sifrai/grandine
NM_DOCKER_TAG=unstable-nethermind-1.35.8
```

**Related issue**
Implements #2373

### DIFF
--- a/nethermind/docker-entrypoint.sh
+++ b/nethermind/docker-entrypoint.sh
@@ -19,6 +19,15 @@ __strip_empty_args() {
 }
 
 
+__normalize_int() {
+  local v=$1
+  if [[ "${v}" =~ ^[0-9]+$ ]]; then
+    v=$((10#${v}))
+  fi
+  printf '%s' "${v}"
+}
+
+
 if [[ -n "${JWT_SECRET}" ]]; then
   echo -n "${JWT_SECRET}" > /var/lib/nethermind/ee-secret/jwtsecret
   echo "JWT secret was supplied in .env"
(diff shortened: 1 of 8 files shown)


nethermind/docker-entrypoint.sh
```diff
@@ -131,9 +140,149 @@ else
   __datadir="--data-dir /var/lib/nethermind"
 fi
 
+if [[ "${COMPOSE_FILE}" =~ grandine-plugin\.yml ]]; then
```
pietjepuk2: Does not match the `allin1` yml. `grandine-plugin.*\.yml`  or `grandine-plugin(-allin1)?\.yml` ?

nethermind/docker-entrypoint.sh
```diff
@@ -131,9 +140,149 @@ else
...
+    fi
+    bootnodes="$(awk -F'- ' '!/^#/ && NF>1 {print $2}' "/var/lib/grandine/testnet/${config_dir}/bootstrap_nodes.yaml" | paste -sd ",")"
+    set +e
+    __grandine+=" --grandine-configurationdirectory=/var/lib/grandine/testnet/${config_dir} --grandine-bootnodes=${bootnodes}"
+  else
+    __grandine+=" --grandine-network=${NETWORK}"
+  fi
+
+  case "${CL_NODE_TYPE}" in
+    archive)
```
pietjepuk2: This becomes a bit messy/repetitive with Grandine stand-alone (aside from the prefix for the options), but.. well, don't really know how to solve it either. Let's hope that we don't get every EL client bundling every CL client 😬

nethermind/docker-entrypoint.sh
yorickdowne: Good catch, thank you!

PR #2424, opened 2026-01-07
Update Vero Dockerfile - remove multiprocessing metrics dir

eth2353:
**What I did**

Synced the Dockerfile with the one upstream - the multiprocessing metrics mode was removed from Vero.

**Related issue**


yorickdowne (approved):
Thanks!

PR #2524, opened 2026-04-10
Support reth download for snapshot

yorickdowne:
**What I did**

Support `reth download` and resolve `static_files` issue, see [Reth issue](https://github.com/paradigmxyz/reth/issues/23436). Requires Reth 2.0.0. This speeds up initial sync dramatically.

Do not offer this for "full" nodes, as Reth has only post-merge, minimal and archive. Use post-merge for post-prague and rolling: This prunes the static files when Reth first runs.

Reth 2.1.0 improves upon this in multiple ways, and the code has comments to show how it should behave after release of 2.1.0

### DIFF
--- a/ethd
+++ b/ethd
@@ -4380,6 +4380,13 @@ __query_4444() {  # Call with with --defaultno for RPC
 }
 
 
+__query_reth_snapshot() {
+  if (whiptail --title "Reth snapshot" --yesno "Do you want to speed up Reth sync by using a database snapshot? This breaks eth_getLogs RPC calls, do not use with RocketPool, SSV or NodeSet" 10 65) then
+    RETH_SNAPSHOT=true
+  fi
+}
+
+
 __query_grafana() {
   if (whiptail --title "Grafana" --yesno "Do you want to use Grafana dashboards?" 10 65) then
       if [[ "$OSTYPE" = "darwin"* ]]; then
@@ -5212,11 +5219,16 @@ config() {
       __query_execution_client
     fi
 
+#shellcheck disable=SC2034
+    RETH_SNAPSHOT=""
     if [[ -n "${EXECUTION_CLIENT+x}" && "${NETWORK}" =~ (mainnet|sepolia) ]]; then
       if [[ "${__deployment}" = "rpc" ]]; then
         __query_4444 --defaultno
       else
         __query_4444 ""
+      fi
+      if [[ "${EXECUTION_CLIENT:-}" = "reth.yml" && "${NETWORK}" = "mainnet" && ! "${EL_NODE_TYPE}" = "full" ]]; then
+        __query_reth_snapshot
       fi
     else  # On all other networks, disable
 #shellcheck disable=SC2034
@@ -5470,7 +5482,9 @@ config() {
   var=MEV_RELAYS
   __update_value_in_env "${var}" "${!var-}" "${__env_file}"
   var=EL_NODE_TYPE
-  __update_value_in_env "${var}" "${!var-false}" "${__env_file}"
+  __update_value_in_env "${var}" "${!var-full}" "${__env_file}"
+  var=RETH_SNAPSHOT
+  __update_value_in_env "${var}" "${!var-}" "${__env_file}"
   var=DOCKER_EXT_NETWORK
   __update_value_in_env "${var}" "${!var:-"rocketpool_net"}" "${__env_file}"
   if [[ "${__deployment}" = "lido_obol" ]]; then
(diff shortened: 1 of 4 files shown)


ethd
```diff
@@ -4380,6 +4380,13 @@ __query_4444() {  # Call with with --defaultno for RPC
 }
 
 
+__query_reth_snapshot() {
+  if (whiptail --title "Reth snapshot" --yesno "Do you want to speed up Reth sync by using a database snapshot? This breaks eth_getLogs RPC calls, do not use with RocketPool, SSV or NodeSet" 10 65) then
```
pietjepuk2, line 4384: With 2.1.0 it will work again with receipts  support?

reth/docker-entrypoint.sh
```diff
@@ -189,6 +200,30 @@ else
...
+  echo "Downloading Reth snapshot with parameters: ${__snap}"
+# Word splitting is desired for the command line parameters
+# shellcheck disable=SC2086
+  reth download --chain "${NETWORK}" --datadir /var/lib/reth ${__static} ${__snap} --resumable
+# Reth 2.1.0
+#  reth download --chain "${NETWORK}" --datadir /var/lib/reth ${__static} ${__snap}
+# Reth 2.0.0 does not take into account --datadir.static, resolve manually
+# Remove with Reth 2.1.0
+  if [[ -n "${__static}" && -d /var/lib/reth/static_files ]]; then
+    mv -v -t /var/lib/static /var/lib/reth/static_files/*
```
pietjepuk2, line 220: If for some reason the `/var/lib/reth/static_files/` is empty, this will fail. Could be fine / intentional, just checking.

reth/docker-entrypoint.sh
```diff
@@ -189,6 +200,30 @@ else
...
+# if [[ ! -d /var/lib/reth/db && "${NETWORK}" = "mainnet" && -n "${RETH_SNAPSHOT}" && ! "${RETH_SNAPSHOT}" = "false" ]]; then
+if [[ ! -d /var/lib/reth/db && ! "${NODE_TYPE}" = "full" && "${NETWORK}" = "mainnet" && -n "${RETH_SNAPSHOT}" \
+      && ! "${RETH_SNAPSHOT}" = "false" ]]; then
+  if [[ ! "${RETH_SNAPSHOT}" = "true" ]]; then
+    __snap="${RETH_SNAPSHOT}"
+  fi
+  echo "Downloading Reth snapshot with parameters: ${__snap}"
+# Word splitting is desired for the command line parameters
+# shellcheck disable=SC2086
+  reth download --chain "${NETWORK}" --datadir /var/lib/reth ${__static} ${__snap} --resumable
```
pietjepuk2, line 214: Does 2.1.0 not have --resumable? Also, if we move the static files, what happens with the actual resumability of this? Are we talking interrupted downloads before we get to the moving?

reth/docker-entrypoint.sh
pietjepuk2, line 96: Could initialize `__snap` here as well for consistency with `__prune`. Or leave as-is, to 'automatically' error out if we ever not initialize it properly in a case branch.

ethd
```diff
@@ -5470,7 +5482,9 @@ config() {
   var=MEV_RELAYS
   __update_value_in_env "${var}" "${!var-}" "${__env_file}"
   var=EL_NODE_TYPE
-  __update_value_in_env "${var}" "${!var-false}" "${__env_file}"
```
pietjepuk2, line 5473: Does that mean people have "EL_NODE_TYPE=false" in their .env? That's not one of our cases (for Reth at least).

ethd
yorickdowne: Yes. There'll be a separate PR for 2.1.0, kept in draft until 2.1.0 is out

reth/docker-entrypoint.sh
yorickdowne: That should not happen: Download is only on a fresh sync, and files are extracted into static_files. If the download fails to the point where no static files are extracted, the -d check should fail.

reth/docker-entrypoint.sh
yorickdowne: 2.1.0 makes `--resumable` the default

Yes, this resumes during this command, not on re-execution. So download, resume stalled files (typically state), download will also extract. Then move if we have an ANCIENT_DIR.

reth/docker-entrypoint.sh
yorickdowne: I'd rather error out. There is no good "default" snap download, only the ones that are right for each node type

ethd
yorickdowne: It means this set EL_NODE_TYPE to false before if EL_NODE_TYPE was not set at all. A relic from when it was EL_MINIMAL_NODE

pietjepuk2 (approved):

PR #2525, opened 2026-04-12
Offer Web3signer during ./ethd config

yorickdowne:
**What I did**

Offer Web3signer during a fresh `./ethd config`, if a VC that supports Web3signer is being used. Deliberately does not offer this for SSV or Obol.

### DIFF
--- a/ethd
+++ b/ethd
@@ -4398,6 +4398,16 @@ __query_reth_snapshot() {
 }
 
 
+__query_web3signer() {
+  if [[ "${__minty_fresh}" -eq 0 ]]; then  # Do not force a key migration, only query on fresh install
+    return
+  fi
+  if (whiptail --title "Web3signer" --yesno "Do you want to keep validator keys in Web3signer?" 10 65) then
+    WEB3SIGNER=true
+  fi
+}
+
+
 __query_grafana() {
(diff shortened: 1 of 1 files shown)


ethd
```diff
@@ -5224,6 +5237,7 @@ config() {
       __query_execution_client
       if [[ ! "${__deployment}" =~ ^(ssv|lido_ssv|rpc)$  ]]; then
         __query_validator_client
+        __query_web3signer
```
pietjepuk2, line 5249: This would cause a double prompt for web3signer right?  Also would prompt if lido_obol?

ethd
```diff
@@ -5149,18 +5159,21 @@ config() {
 
   __run_pre_update_script
 
+  WEB3SIGNER=false
```
pietjepuk2: Should we not only set it to false when it's __minty_fresh=1 ? Otherwise get it from the config?

ethd
yorickdowne: Yeah it would. Fixed that.

tbh the logic flow for config() is an unholy mess. It needs to be refactored. But not in this PR.

ethd
yorickdowne: You are absolutely right. Fixed that.

pietjepuk2 (approved):

PR #2527, opened 2026-04-12
Improve Reth snapshot download for Reth 2.1.0

yorickdowne:
**What I did**

`--resumable` is now default, receipts and bodies download can be specified on CLI

Keep in draft until Reth 2.1.0 is out

### DIFF
--- a/ethd
+++ b/ethd
@@ -4466,14 +4466,13 @@ __query_4444() {  # Call with with --defaultno for RPC
 
 
 __query_reth_snapshot() {
-  # Reth 2.1.0, remove the test for full
-  if [[ "${EXECUTION_CLIENT}" != "reth.yml" || "${NETWORK}" != "mainnet" || "${EL_NODE_TYPE}" = "full" ]]; then
+  if [[ "${EXECUTION_CLIENT}" != "reth.yml" || "${NETWORK}" != "mainnet" ]]; then
     return
   fi
 
   __write_vars+=("RETH_SNAPSHOT")
 
-  if whiptail --title "Reth snapshot" --yesno "Do you want to speed up Reth sync by using a database snapshot? This breaks eth_getLogs RPC calls, do not use with RocketPool, SSV or NodeSet" 10 65; then
+  if whiptail --title "Reth snapshot" --yesno "Do you want to speed up Reth sync by using a database snapshot?" 10 65; then
     RETH_SNAPSHOT=true
   else
     RETH_SNAPSHOT=""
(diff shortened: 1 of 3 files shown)


yorickdowne:
See https://github.com/paradigmxyz/reth/pull/23513 for reference

reth/docker-entrypoint.sh
```diff
@@ -214,14 +206,7 @@ if [[ -n "${__snap}" && ! -d /var/lib/reth/db ]]; then
...
 # shellcheck disable=SC2086
-  reth download --chain "${NETWORK}" --datadir /var/lib/reth ${__static} ${__snap} --resumable
-# Reth 2.1.0
-#  reth download --chain "${NETWORK}" --datadir /var/lib/reth ${__static} ${__snap}
-# Reth 2.0.0 does not take into account --datadir.static, resolve manually
-# Remove with Reth 2.1.0
-  if [[ -n "${__static}" && -d /var/lib/reth/static_files ]]; then
-    mv -v -t /var/lib/static /var/lib/reth/static_files/*
-  fi
+  exec reth download --chain "${NETWORK}" --datadir /var/lib/reth ${__static} ${__snap}
```
pietjepuk2, line 209: I assume the "exec" is deliberate? Just like e.g. prune

ethd
```diff
@@ -4466,14 +4466,13 @@ __query_4444() {  # Call with with --defaultno for RPC
...
-  # Reth 2.1.0, remove the test for full
-  if [[ "${EXECUTION_CLIENT}" != "reth.yml" || "${NETWORK}" != "mainnet" || "${EL_NODE_TYPE}" = "full" ]]; then
+  if [[ "${EXECUTION_CLIENT}" != "reth.yml" || "${NETWORK}" != "mainnet" ]]; then
     return
   fi
 
   __write_vars+=("RETH_SNAPSHOT")
 
-  if whiptail --title "Reth snapshot" --yesno "Do you want to speed up Reth sync by using a database snapshot? This breaks eth_getLogs RPC calls, do not use with RocketPool, SSV or NodeSet" 10 65; then
+  if (whiptail --title "Reth snapshot" --yesno "Do you want to speed up Reth sync by using a database snapshot?" 10 65) then
```
pietjepuk2: From what I can tell, we do `if whiptail` everywhere, not `if (whiptail`. Might be worth keeping consistency.

reth/docker-entrypoint.sh
yorickdowne: Yes it's deliberate. If a user changes their mind and wants to quickly resync again, the SIGINT/SIGTERM makes it to Reth, instead of having to wait 5 mins for force kill

ethd
yorickdowne: Yes it is. Dang how did that sneak in. One sec.

reth/docker-entrypoint.sh
pietjepuk2: Clear!

pietjepuk2 (approved):

PR #2532, opened 2026-04-14
Reth snapshot download allows any network w/ custom params

yorickdowne:
**What I did**

A user wanted to use `--full --url <pandaops>` on Hoodi, and the entrypoint doesn't allow downloads on Hoodi.

Change the entrypoint so users can deploy footguns to the max. The one above actually works.

### DIFF
--- a/reth/docker-entrypoint.sh
+++ b/reth/docker-entrypoint.sh
@@ -167,6 +167,15 @@ case "${NODE_TYPE}" in
     ;;
 esac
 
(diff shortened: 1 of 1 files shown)


reth/docker-entrypoint.sh
pietjepuk2, line 217: Definitely a foot gun with custom networks and all, but by default it should still work (even with custom networks). Only if users start combining RETH_SNAPSHOT with `NETWORK=https://..` will there be issues. But as you said, footgun is allowed.

pietjepuk2 (approved):

PR #2535, opened 2026-04-17
Alloy

yorickdowne:
**What I did**

Promtail is EOL since March 2nd 2026, replace with alloy

Log collection targets are now discovered via labels

Metrics collection is done by Alloy

Keep node-exporter and cadvisor, they are fairly involved with the host and not easy to fully reproduce in alloy

`grafana-rootless.yml` now scrapes Docker socket as well. This has been tested and works.

`ethd` detects rootless and sets `DOCKER_SOCK` accordingly. This has been tested and works.

This is a breaking change for users of `custom-prom.yml` and `custom-lokiurl.yml`. Release notes will point this out, and users can with minimal effort recreate their remote write targets in alloy. If they have custom scrapes in `./prometheus/conf.d`, those can be converted to alloy format by alloy, which will also be documented in release notes.

These are advanced, niche uses, and I would not be surprised if there are no users of those features other than the Grandine team, who have been made aware. The features were created for CryptoManufaktur, and we moved away from that to "central-proxy-docker" a few years ago.


tempo-traefik.yml
```diff
@@ -0,0 +1,16 @@
+# To be used in conjunction with grafana.yml, for trace collection
+# Please be very cautious, Loki is not authenticated by default
+services:
+  loki:
```
pietjepuk2: Probably a copy/pasta mistake. Needs to be `tempo:` I assume

alloy-obol/docker-entrypoint.sh
```diff
@@ -0,0 +1,41 @@
...
+  CLUSTER_NAME=dummy-cluster
+fi
+
+if [ -z "${CLUSTER_PEER:-}" ]; then
+  echo "Error: \$CLUSTER_PEER variable is empty" >&2
+  CLUSTER_PEER=dummy-peer
+fi
+
+if [ -z "${OBOL_PROM_REMOTE_WRITE_TOKEN:-}" ]; then
+  echo "Error: \$ OBOL_PROM_REMOTE_WRITE_TOKEN variable is empty" >&2
```
pietjepuk2: `\$OBOL_PROM_REMOTE_WRITE_TOKEN`  (remove stray space)

grafana-cloud.yml
```diff
@@ -138,28 +115,32 @@ services:
...
-    user: root
+  alloy:
+    image: grafana/alloy:latest
+    restart: unless-stopped
     volumes:
-      - /etc/machine-id:/etc/machine-id:ro
-      - ./promtail:/etc/promtail
-      - promtail-data:/tmp
+      - alloy-data:/var/lib/alloy
+      - ./alloy:/etc/alloy
```
pietjepuk2, line 125: This doesn't discover files in the config.d dir I think.  Maybe add it explicitly as well?

> If you provide a directory path for the <PATH_NAME>, Alloy finds *.alloy files, ignoring nested directories, and loads them as a single configuration source. However, component names must be unique across all Alloy configuration files, and configuration blocks must not be repeated.

( https://grafana.com/docs/alloy/latest/reference/cli/run/ )

tempo-traefik.yml
yorickdowne: Yes it does

grafana-cloud.yml
yorickdowne: Foiled by AI, I didn't test that fully and it told me it'd be nested. So, alloy can't take two arguments. I tried. I removed config.d and users will just stick their stuff straight into `./alloy`

pietjepuk2 (approved):

PR #2548, opened 2026-04-20
ethd works better for users without sudo

yorickdowne:
**What I did**

`ethd install` run as `alice` with Eth Docker owned by `node` will now correctly add `alice` to the `docker` group, not `node`; whether `alice` runs `ethd install` or `sudo ethd install`.

`chown` and `chmod` on config file locations work when `alice` runs `ethd` and `node:node-admins` owns `eth-docker`

In several locations, check that the user can `sudo` before invoking `${__auto_sudo}`

Two small style fixes

### DIFF
--- a/ethd
+++ b/ethd
@@ -279,12 +279,12 @@ __upgrade_docker() {
               echo "Docker ${__docker_version} with runc ${runc_version} detected"
               echo "This version of runc is vulnerable"
               __nag_os_version
-              if [ "${__eol_os}" -eq 1 ]; then
+              if [[ "${__eol_os}" -eq 1 ]]; then
                 echo "${__project_name} cannot update runc on Ubuntu ${__os_major_version}."
                 return
               fi
               echo "It is recommended that you update runc"
-              if [ "${__non_interactive}" -eq 0 ]; then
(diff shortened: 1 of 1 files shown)


ethd
```diff
@@ -2285,7 +2291,7 @@ __verify_ancient_dir() {
     return 1
   fi
 
-  if ! mapfile -t entries < <(${__auto_sudo} find "${dir}" -mindepth 1 -maxdepth 1 -exec basename {} \;); then
+  if [[ "${__cannot_sudo}" -eq 0 ]] && ! mapfile -t entries < <(${__auto_sudo} find "${dir}" -mindepth 1 -maxdepth 1 -exec basename {} \;); then
```
pietjepuk2, line 2294: Should the function not also `return 1` if `"${__cannot_sudo}" -eq 1`? Now it returns as if listing is empty

yorickdowne: The intent is to skip this check. This is a niche check, it checks whether `ANCIENT_DIR` ended up as a file. Extremely unlikely.

If a user can't sudo, don't fail, just trust the `ANCIENT_DIR` is indeed a directory.

pietjepuk2 (approved):

PR #2556, opened 2026-04-21
Use UBUNTU_CODENAME if defined

yorickdowne:
**What I did**

When installing docker-ce, use `${UBUNTU_CODENAME:-VERSION_CODENAME}` for the `Suites:`. `UBUNTU_CODENAME` if defined, `VERSION_CODENAME` otherwise. Works well on Ubuntu and Debian.

### DIFF
--- a/ethd
+++ b/ethd
@@ -836,7 +836,7 @@ __install_docker() {
     ${__auto_sudo} tee /etc/apt/sources.list.d/docker.sources <<EOF
 Types: deb
 URIs: https://download.docker.com/linux/${repo}
-Suites: $(. /etc/os-release && echo "$VERSION_CODENAME")
+Suites: $(. /etc/os-release && echo "${UBUNTU_CODENAME:-${VERSION_CODENAME}}")
 Components: stable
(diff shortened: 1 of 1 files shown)


ethd
```diff
@@ -787,7 +787,7 @@ __install_docker() {
     ${__auto_sudo} tee /etc/apt/sources.list.d/docker.sources <<EOF
 Types: deb
 URIs: https://download.docker.com/linux/${repo}
-Suites: $(. /etc/os-release && echo "$VERSION_CODENAME")
+Suites: $(. /etc/os-release && echo "${UBUNTU_CODENAME:-VERSION_CODENAME}")
```
pietjepuk2: `${VERSION_CODENAME}` , otherwise it just takes the string as-is no?

pietjepuk2 (changes requested):

ethd
yorickdowne: Yes indeed. Complete oversight. Thanks for catching

yorickdowne: Fixed

pietjepuk2 (approved):

PR #2558, opened 2026-04-22
Invalidate build cache less often

yorickdowne:
**What I did**

Users that `./ethd update` frequently get to keep their build cache, until 50GB are reclaimable. Strike a balance between cleaning up Docker artefacts and constantly rebuilding from scratch.

Parsing the `docker system df` output is brittle. There are multiple safeguards in place to just return and do nothing, instead of failing `ethd` entirely.

### DIFF
--- a/ethd
+++ b/ethd
@@ -1943,13 +1943,68 @@ __nag_os_update() {
 }
 
 
+__docker_df_to_bytes() {
+  local awk_exe
+
+  if [[ "$OSTYPE" = "darwin"* ]]; then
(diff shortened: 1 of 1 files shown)


yorickdowne:
Putting this to draft. I'm worried it'll keep the build from pulling in changed `latest` clients. Need to test more

yorickdowne:
Out of draft. I `build --pull` which makes it OK. Specifically in `./ethd update`. A user that runs `./ethd cmd build` may miss an update, but if they run `./ethd cmd build --pull` they will not.

pietjepuk2 (approved):

PR #2559, opened 2026-04-22
Support Web3signer distroless

yorickdowne:
**What I did**

Consensys introduced a `latest-distroless` tag. This supports it.

Requires Web3signer 26.4.2 or later

The migration warnings and failures have migrated into `w3s-init`

`Dockerfile.binary` is the default and will just run the entrypoint of whatever image the user chooses

`Dockerfile.source` still supports custom network processing

Add Web3signer CI

### DIFF
--- a/.github/workflows/test-web3signer.yml
+++ b/.github/workflows/test-web3signer.yml
@@ -1,0 +1,80 @@
+name: Test Web3signer
+
+defaults:
+  run:
+    shell: bash
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+    types: [opened, synchronize, labeled, unlabeled]
+    branches: [main]
+
+concurrency:
+  group: ci-${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  test-web3signer:
+    if: |
+      contains(github.event.pull_request.labels.*.name, 'test-web3signer') ||
+      contains(github.event.pull_request.labels.*.name, 'test-all') ||
+      github.event_name == 'push'
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v6
+      - name: Set up Docker buildx
+        uses: docker/setup-buildx-action@v4
+      - name: Create .env file
+        run: cp default.env .env
+      - name: Set Lighthouse/Geth/Web3signer
+        run: |
+          source ./.github/helper.sh
+          CORE_FILES=lighthouse.yml:geth.yml:web3signer.yml
+          var=CORE_FILES
+          set_value_in_env
+          FEE_RECIPIENT=0xDccf8451070a86183eE70D330C4c43b686E9CF86
+          var=FEE_RECIPIENT
+          set_value_in_env
+          WEB3SIGNER=true
+          var=WEB3SIGNER
+          set_value_in_env
+          W3S_DOCKER_TAG=latest
+          var=W3S_DOCKER_TAG
+          set_value_in_env
+          W3S_READ_ONLY=false
+          var=W3S_READ_ONLY
+          set_value_in_env
+      - name: Start Lighthouse/Geth/Web3signer
+        run: ./ethd up
+      - name: Pause for 30 seconds
+        run: sleep 30
+      - name: Test Web3signer
+        run: ./.github/check-service.sh web3signer
+      - name: Test PostgreSQL
+        run: ./.github/check-service.sh postgres
+      - name: Test version
+        run: ./ethd version
+      - name: Set distroless
+        run: |
+          source ./.github/helper.sh
+          W3S_DOCKER_TAG=latest-distroless
+          var=W3S_DOCKER_TAG
+          set_value_in_env
+          W3S_READ_ONLY=true
+          var=W3S_READ_ONLY
+          set_value_in_env
+      - name: Rebuild Web3signer image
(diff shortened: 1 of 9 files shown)


.github/workflows/test-web3signer.yml
```diff
@@ -0,0 +1,77 @@
...
+          var=WEB3SIGNER
+          set_value_in_env
+          W3S_DOCKER_TAG=latest-distroless
+          var=W3S_DOCKER_TAG
+          set_value_in_env
+          W3S_READ_ONLY=true
+          var=W3S_READ_ONLY
+          set_value_in_env
+      - name: Start Lighthouse/Geth/Web3signer
+        run: ./ethd up
```
pietjepuk2, line 72: This only 'recreates', but does not rebuild with the 'latest-distroless' tag. 

Needs a 
```
- name: Rebuild Web3signer distroless images
          run: docker compose build --pull web3signer w3s-init
```

(Codex spotted this)

default.env
```diff
@@ -404,8 +404,13 @@ VERO_DOCKER_REPO=ghcr.io/serenita-org/vero
 VERO_DOCKERFILE=Dockerfile.binary
 
 # Web3Signer
+# Use latest-distroless or <vx.y>-distroless for a distroless setup
 W3S_DOCKER_TAG=latest
+# Set to "true" for distroless and read-only Docker
+W3S_READ_ONLY=false
 W3S_DOCKER_REPO=consensys/web3signer
+# Use Dockerfile.custom-network for a custom devnet NETWORK. Doesn't support distroless.
+W3S_DOCKERFILE=Dockerfile.slim
```
pietjepuk2: Probably rather uncommon that people have custom networks. But this will break those setups right (until they explicitly change this value to `Dockerfile.custom-network`)

ethd
```diff
@@ -5,7 +5,7 @@ set -Eeuo pipefail
 __project_name="Eth Docker"
 __app_name="Ethereum node"
```
pietjepuk2, line 6: in version()

`__docompose exec web3signer /opt/web3signer/bin/web3signer --version` does not work with distroless. Not sure what the best approach is. Can't have helper scripts, because distroless. So maybe just `docker run` for this case?

web3signer.yml
```diff
@@ -42,13 +43,14 @@ services:
...
     image: web3signer:local
     pull_policy: never
-    user: web3signer
+    user: 10000:10000
     volumes:
       - web3signer-keys:/var/lib/web3signer
       - /etc/localtime:/etc/localtime:ro
+    read_only: ${W3S_READ_ONLY:-false}
     environment:
       - JAVA_OPTS=${W3S_HEAP:--Xmx6g}
```
pietjepuk2: Apparently this does not work with 'distroless' (according to Codex at least).

The web3signer script in the original image did

```
eval "set -- $(
          printf '%s\n' "$DEFAULT_JVM_OPTS $JAVA_OPTS $WEB3SIGNER_OPTS" |
          xargs -n1 |
          sed ' s~[^-[:alnum:]+,./:=@_]~\\&~g; ' |
          tr '\n' ' '
      )" '"$@"'

  exec "$JAVACMD" "$@"
```

But distroless does not. 

Can set `JAVA_TOOL_OPTIONS` instead, but... only for distroless? Or also for regular image?

.github/workflows/test-web3signer.yml
yorickdowne: That's in CI. The image hasn't changed since 30s ago when we ran without W3S_READ_ONLY. We do not need to rebuild.

default.env
yorickdowne: Correct. I had a choice of breaking it entirely or gating it behind this change. I asked Consensys whether they use the feature. They do not. It's a feature that Pandaops had asked for on behalf of client teams, and I am now unsure whether any client teams use it.

It'd be used for stuff like "devnet3"

web3signer.yml
yorickdowne: `read_only` is a Docker option. This absolutely works. So well in fact they had to fix Web3signer-distroless because it'd break when I turned that on: https://github.com/Consensys/web3signer/issues/1175

ethd
yorickdowne: Very good point. Fixed.

default.env
yorickdowne: On further thought. You likely can't even run a custom devnet with an image, it'd need to be source-built. Which means no one is using this feature

.github/workflows/test-web3signer.yml
yorickdowne: Yep I was wrong about that. Fixed

web3signer.yml
yorickdowne: AH ok, JAVA_OPTS. Yeah this likely doesn't work in distroless. We can point that out to Consensys as a bug

yorickdowne: Let's leave that for another PR. `JAVA_TOOL_OPTIONS` is a good choice; we'd need to verify it doesn't interfere with what regular web3signer is doing

default.env
yorickdowne: I've added a Dockerfile.source for this use case

web3signer.yml
yorickdowne: On further thought, `JAVA_TOOL_OPTIONS` it is

pietjepuk2 (approved):

PR #2561, opened 2026-04-22
Add command to reduce security of web3signer keys

yorickdowne:
**What I did**

Add a command to reduce the security of validator key files in Web3signer, to load in seconds, not 30 minutes.

This comes in handy when running genesis validator keys on  Hoodi. I've done it manually one too many times, and wanted a command.

Uses https://github.com/usmansaleem/v4keystore_converter

Will refuse to run on `mainnet` or `gnosis`

### DIFF
--- a/web3signer/convert-keys.sh
+++ b/web3signer/convert-keys.sh
@@ -1,0 +1,56 @@
+#!/usr/bin/env bash
+# shellcheck disable=SC2174
+set -Eeuo pipefail
+
+ts=$(date +%s)
+base_dir=/var/lib/web3signer
+mkdir -p -m 700 "${base_dir}"/converted-keys
+
+if ! find "${base_dir}"/keys -type f -name '*.password' -print -quit 2>/dev/null | grep -q .; then
+  echo "No key files found in ${base_dir}/keys. Aborting."
+  exit 1
+fi
+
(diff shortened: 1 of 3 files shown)


web3signer/convert-keys.sh
```diff
@@ -0,0 +1,56 @@
...
+if [[ "${NETWORK}" =~ ^(mainnet|gnosis)$ ]]; then
+  echo "Reducing key security on ${NETWORK} is not recommended. If you need to do so, please do so manually."
+  echo "Aborting"
+  exit 1
+fi
+
+if find "${base_dir}"/converted-keys -type f -name '*.json' -print -quit | grep -q .; then
+  while true; do
+    echo "Keys have previously been converted."
+    read -rp "Do you want to run conversion again, maybe because you added keys? (N/y) " yn
```
pietjepuk2, line 23: Shouldn't we also start clean?  I don't think 'v4keystore_converter' has a resume capability, but of course rerunning it on already reduced-security keys will be fast.

But not starting fresh means we risk copying over stale keys from `converted-keys` to `keys`, that the user already deleted from `keys` at some point.

yorickdowne: That is indeed an excellent point.

pietjepuk2 (approved):

PR #2566, opened 2026-04-23
ethd keys handles malformed error messages

yorickdowne:
**What I did**

Some clients will return a `500` code with a plain-text error, instead of a JSON-formatted one.

Handle error messages gracefully, whether `jq` can parse them, or not.

Also fix debug output of API data that is passed in a file

### DIFF
--- a/vc-utils/keymanager.sh
+++ b/vc-utils/keymanager.sh
@@ -25,6 +25,18 @@ __eth2_val_tools=0
 __non_interactive=0
 
 
+# Not all clients return .message like they should.
+# This is called as __print_jq_message "<message>" "<jq_query>"
+__print_jq_message() {
+  local message=$1
+  local jq_query=$2
+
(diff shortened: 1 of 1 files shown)


vc-utils/keymanager.sh
```diff
@@ -25,6 +25,18 @@ __eth2_val_tools=0
 __non_interactive=0
 
 
+# Not all clients return .message like they should.
+# This is called as __print_jq_message "<message>" "<jq_query>"
+__print_jq_message() {
+  local message=$1
+  local jq_query=$2
+
+  if ! echo "${message}" | jq -r "${jq_query}" 2>/dev/null; then
```
pietjepuk2: Should be `jq -er "${jq_query} // empty"`  (e for more strict, // empty to not also print the null)

yorickdowne: Tested that, 100% right. This also means some other parts of Eth Docker that use this pattern need to be adjusted

pietjepuk2 (approved):

PR #2567, opened 2026-04-24
Adjust .motd to actual directory

yorickdowne:
**What I did**

When a user wants to call `ethd` from a `~/.profile` alias, adjust `.motd` to mention the actual installation directory


ethd
```diff
@@ -473,6 +473,10 @@ __check_compose_version() {
 
 
 __prep_conffiles() {
+# Create .motd if it doesn't exist
+  if [[ ! -f ./.motd ]]; then
+    ${__as_owner} cp ./.motd.sample ./.motd
```
pietjepuk2, line 478: Existing installs will just copy this sample verbatim, without the "sed" stuff you added below. Is that intended?

yorickdowne: Yes that’s intended. This only gets adjusted during install

pietjepuk2 (approved):

PR #2568, opened 2026-04-24
Support Nethermind FlatDB

yorickdowne:
**What I did**

New `.env` variable `NETHERMIND_FLATDB`. Requires Nethermind 1.37.1

- Offer Flat and FlatInTrie
~~- Auto-convert existing DBs~~
- Ignore FlatDB for archive nodes
- Do not configure state pruning when using FlatDB

### DIFF
--- a/default.env
+++ b/default.env
@@ -258,6 +258,9 @@ EL_NODE_TYPE=pre-merge-expiry
 # empty or "false" disables it, and any other string is assumed to be
 # exact parameters for Reth's "download" function
 RETH_SNAPSHOT=
+# Nethermind FlatDB. Possible values empty (don't use FlatDB), flat, or flatintrie, see https://github.com/NethermindEth/nethermind/releases/tag/1.37.0
+# FlatDB cannot be used with an archive node, and Flat requires a node with 48GB of RAM. FlatInTrie is less memory intensive.
+NETHERMIND_FLATDB=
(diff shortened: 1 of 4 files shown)


yorickdowne:
Wait for 1.37.1 for a FlatDB sync with Nimbus fix

yorickdowne:
Removing always-on migration parameter, it does not work well with a restart during sync. See https://github.com/NethermindEth/nethermind/issues/11418

default.env
```diff
@@ -258,6 +258,9 @@ EL_NODE_TYPE=pre-merge-expiry
 # empty or "false" disables it, and any other string is assumed to be
 # exact parameters for Reth's "download" function
 RETH_SNAPSHOT=
+# Nethermind FlatDB. Possible values empty (don't use FlatDB), flat, or flatintrie, see https://github.com/NethermindEth/nethermind/releases/tag/1.37.0
+# FlatDB cannot be used with an archive node.
```
pietjepuk2, line 262: Am I correct in understanding that switching means a fresh sync?  `--FlatDb.ImportFromPruningTrieState=true` is commented out?

yorickdowne: Correct, there is a bug where restarting the sync triggers the import. Restarting the sync is generally not working right now - that aside, this remains disabled until https://github.com/NethermindEth/nethermind/issues/11418 has been addressed and that's in a release.

pietjepuk2 (approved):

PR #2569, opened 2026-04-25
Warn user if their Docker-CE is old

yorickdowne:
**What I did**

A user mentioned that they'd run for a year+ with an old Docker-CE because they did not re-enable the repo after Ubuntu upgrade. Detect and warn. It's a bit manual, by design: When Docker-CE 30 is stable, maybe around 30.3, `ethd` needs to be told, and so on.

I am also deliberately not chasing minor versions. Just check that the user ran an upgrade sometime recently-ish, and their major version is current.

### DIFF
--- a/ethd
+++ b/ethd
@@ -6,9 +6,13 @@ __project_name="Eth Docker"
 __app_name="Ethereum node"
 __sample_service="consensus"
 __min_env_version=55
+__target_pg=18
+__current_docker=29
 __docker_exe="docker"
 __old_docker=0
+__oldish_docker=0
 __docker_sudo=""
+__docker_version=""
 __docker_major_version=""
 __docker_minor_version=""
 __docker_patch_version=""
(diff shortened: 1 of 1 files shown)


ethd
```diff
@@ -5,10 +5,14 @@ set -Eeuo pipefail
 __project_name="Eth Docker"
 __app_name="Ethereum node"
 __sample_service="consensus"
-__min_env_version=56
+__min_env_version=55
```
pietjepuk2: Probably unintented left over from rebasing?

ethd
```diff
@@ -6358,3 +6367,15 @@ if [[ "${__old_compose}" -eq 1 && "${__compose_major}" -eq 2 ]]; then
     echo "Please do so by running: \"sudo apt update && sudo apt dist-upgrade\""
   fi
 fi
+
+if [[ "${__oldish_docker}" -eq 1 ]]; then
+  echo "You are using Docker-CE version ${__docker_version}, which is not current."
+  current_docker_release=$(curl -m5 -s -H "User-Agent: eth-docker" "https://api.github.com/repos/moby/moby/releases/latest" 2>/dev/null | jq -r '.tag_name // empty' 2>/dev/null) || true
```
pietjepuk2: This returns `docker-v29.4.2` for me right now.    Using `.name` instead does return something starting with `v`, which is what the below regex match needs

ethd
yorickdowne: Yep left over, fixing

ethd
yorickdowne: Ah good point. So it'd never print that. Changing.

pietjepuk2 (approved):

PR #2571, opened 2026-04-25
Fix __owner_group for non-setgid setups

### LINKED ISSUE #2570: Spurious ownership fix on every start when directory group differs from owner's primary group
## Description

Commit 7c07d3a ("Make it easier to run eth-docker with an app user", PR #2545) changed `__owner_group` from `id -gn "${__owner}"` (owner's primary group) to `ls -ld . | awk '{print $4}'` (directory's group).

The PR's intent was to support a multi-admin setup: an app user `node` owns eth-docker, a group like `node-admins` has directory ownership with **setgid** set, and multiple admins can manage the node. With setgid, new files inherit the directory's group, so using the directory group is correct.

However, in the common single-user setup **without setgid**, new files inherit the creator's **primary group**, not the directory group. When these differ (e.g., directory is `cthulhu:docker` but primary group is `cthulhu`), the ownership check in `__prep_conffiles` sees a perpetual mismatch and "fixes" `.env` on every start, producing spurious "Fixing ownership" messages.

## Steps to Reproduce

1. Have a user whose primary group differs from the directory's group (e.g., directory owned by `cthulhu:docker`, user's primary group is `cthulhu`)
2. Run `./ethd up` or any command that triggers `__prep_conffiles`
3. Observe "Fixing ownership" messages on every invocation, even though ownership is correct

## Expected Behavior

No "Fixing ownership" message when the files already have the correct owner and group.

## Proposed Fix

Check whether the setgid bit is set on the directory:
- If setgid is active, use the directory's group (preserving the multi-admin behavior from PR #2545)
- If setgid is not active, use the owner's primary group (the pre-regression behavior)

f-barth:
## Summary

- Fix regression from commit 7c07d3a (PR #2545) where `__owner_group` always uses the directory's group, causing spurious "Fixing ownership" messages on every start when the directory group differs from the owner's primary group
- Check the setgid bit on the directory: use directory group when set (multi-admin setup), owner's primary group otherwise (standard single-user setup)

## Test plan

- [ ] On a system **without** setgid on the eth-docker dir (the common case): verify `__owner_group` equals the owner's primary group and no spurious "Fixing ownership" messages appear
- [ ] On a system **with** setgid (app-user setup from PR #2545): verify `__owner_group` equals the directory's group, preserving multi-admin behavior

Fixes #2570

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/ethd
+++ b/ethd
@@ -6061,8 +6061,14 @@ cd "$(dirname "$(realpath "${BASH_SOURCE[0]}")")"
 # Use this to make sure root doesn't end up owning files
 # shellcheck disable=SC2012
 __owner=$(ls -ld . | awk '{print $3}')
-# shellcheck disable=SC2012
-__owner_group=$(ls -ld . | awk '{print $4}')
+if [[ -g . ]]; then
+  # setgid: new files inherit directory group
+  # shellcheck disable=SC2012
+  __owner_group=$(ls -ld . | awk '{print $4}')
+else
+  # no setgid: new files inherit owner's primary group
+  __owner_group=$(id -gn "${__owner}")
+fi
 
 if [[ "${__owner}" = "root" ]]; then
   echo "Please install ${__project_name} as a non-root user."
--- a/ethd
+++ b/ethd
@@ -6060,11 +6060,11 @@ fi
 cd "$(dirname "$(realpath "${BASH_SOURCE[0]}")")"
 # Use this to make sure root doesn't end up owning files
 # shellcheck disable=SC2012
-__owner=$(ls -ld . | awk '{print $3}')
+__dir_listing=$(ls -ld .)
+__owner=$(awk '{print $3}' <<< "${__dir_listing}")
 if [[ -g . ]]; then
   # setgid: new files inherit directory group
-  # shellcheck disable=SC2012
-  __owner_group=$(ls -ld . | awk '{print $4}')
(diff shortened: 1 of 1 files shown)


yorickdowne:
I appreciate the thought, Florian, I do. But this does not express the intent. For example, if the directory is `user:admin-group` and the user is `user:user`, and there is not setgid, then your code would cause an adjustment of `dkg_output` to `user:user`, which is not desirable:

```
if find .eth/dkg_output \( \! -user "${__owner}" -o \! -group "${__owner_group}" \) -print -quit | grep -q .; then
    if [[ "${__cannot_sudo}" -eq 0 ]]; then
      echo "Fixing ownership of .eth/dkg_output"
      ${__auto_sudo} chown -R "${__owner}:${__owner_group}" .eth/dkg_output
      ${__as_owner} chmod u=rwx,go=rx .eth/dkg_output
    else
      echo "Ownership of .eth/dkg_output should be fixed, but this user can't sudo"
    fi
  fi
```
  
I am currently iterating on this multi-user code in  #2575 , and just handling the check for whether `.env` needs to be adjusted better.

Give me a bit of time to test that thoroughly, and for pietjepuk to review it when it's ready. In the meanwhile, please `chown -R user:user <eth-docker-dir>` and don't do a `user:admin-group` setup quite yet.

I'll close this PR once I have something that works in a multi-user setup *and* doesn't break the typical user:user setup

PR #2572, opened 2026-04-25
Update to latest tagged release instead of tracking main

f-barth:
## Summary

- Default updates (`ETH_DOCKER_TAG` empty or `latest`) now fetch tags and check out the most recent release instead of `git pull origin main`
- Pinned updates (`ETH_DOCKER_TAG` set to a specific version) are unchanged
- Falls back to pulling `main` if no tags exist

## Motivation

Production staking systems benefit from running tested, tagged releases rather than tracking `main`. Currently, `ethd update` pulls whatever is on `main` at the time, which may include untested or in-progress changes. By defaulting to the latest tagged release, operators get stable versions that have been explicitly cut as releases.

The existing `ETH_DOCKER_TAG` pinning mechanism already uses this exact pattern (fetch tags, checkout on a local `tag-*` branch) — this change simply applies it to the default path as well, automatically selecting the latest tag via `git tag --sort=-v:refname`.

## Test plan

- [ ] `ethd update` with `ETH_DOCKER_TAG=` (default): should fetch tags and check out the latest one (currently `v26.4.1`)
- [ ] `ethd update` with `ETH_DOCKER_TAG=v26.3.0`: should still pin to that specific tag (unchanged behavior)
- [ ] Edge case: repo with no tags falls back to `git pull origin main`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/ethd
+++ b/ethd
@@ -2149,11 +2149,14 @@ update() {
     __get_value_from_env "${var}" "${__env_file}" "__value"
     if [[ -z "${__value}" || "${__value}" = "latest" ]]; then
       export ETHDPINNED=""
-      branch=$(git rev-parse --abbrev-ref HEAD)
-      if [[ "${branch}" =~ ^tag-* ]]; then
-        git checkout main
+      ${__as_owner} git fetch origin --tags
+      __value=$(git tag --sort=-v:refname | head -1)
+      if [[ -z "${__value}" ]]; then
+        echo "No tags found, falling back to main branch"
+        ${__as_owner} git pull origin main
+      else
+        ${__as_owner} git checkout -B "tag-${__value}" "tags/${__value}"
       fi
-      ${__as_owner} git pull origin main
     else
       export ETHDPINNED="${__value}"
       ${__as_owner} git fetch --tags
--- a/ethd
+++ b/ethd
@@ -2150,9 +2150,10 @@ update() {
     if [[ -z "${__value}" || "${__value}" = "latest" ]]; then
       export ETHDPINNED=""
       ${__as_owner} git fetch origin --tags
-      __value=$(git tag --sort=-v:refname | head -1)
+      __value=$(${__as_owner} git tag --sort=-v:refname | head -1)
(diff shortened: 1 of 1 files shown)


yorickdowne:
"Production staking systems benefit from running tested, tagged releases rather than tracking main."

Agreed. That is what `ETH_DOCKER_TAG` is for. Production staking systems also tag client releases, and blindly pulling an Eth Docker release *will* break things when Eth Docker relies on the behavior of a new client release.

Either
- Tag all client releases and eth docker itself; read all release notes before deciding to bump tag
OR
- Don't tag eth docker or client releases (latest, stable), `./ethd update` pulls something in that's very likely to work with all-updated releases

This half-way-between idea would, I fear, make things worse

yorickdowne:
On further thought. Can you change your code so it picks up on `ETH_DOCKER_TAG=stable`. Grabs the tag that Github thinks is latest with something like `latest_tag=$(curl -s https://api.github.com/repos/ethstaker/eth-docker/releases/latest | jq -r .tag_name)` - the host might not have curl or jq, check for that and install as needed - then sets the `__value` to that tag and continues as now from there.

Document the tag in default.env, keep `ETH_DOCKER_TAG=` as the default and make clear that the default is to update the code to the current development state.

yorickdowne:
This is a bit of an issue: `__value=$(${__as_owner} git tag --sort=-v:refname | head -1)` .. this would also pick up on pre-release tags, and that's likely not your intent

That's why the curl idea. There might be a better way tbh - but `latest` release is not a git feature, it's a github feature

yorickdowne:
Closing in favor of https://github.com/ethstaker/eth-docker/pull/2596 . Thank you for the idea. I believe `ETH_DOCKER_TAG=stable` will give you what you are looking for. Please beware of interactions with pinned client versions, or of what happens if a client removes a parameter that Eth Docker relies upon, with that change being in `main` but not yet a release. Erigon used to be notorious for that, for example.

PR #2573, opened 2026-04-25
Deal with umask 077

yorickdowne:
**What I did**

A user ran with `umask 077`, which wreaks merry havoc with permissions when bind-mounting

Fix what I am aware of. This may not be comprehensive. deposit-cli has an argument to get the uid, and keymanager runs as root. ethdo bind-mounts .eth as well


yorickdowne:
This needs to become more generic - iterate through a dictionary or something, so I can fix all directories and files without repeating the code over and over. And, that's for another PR

PR #2575, opened 2026-04-26
Better handling for a multi-user setup

### LINKED ISSUE #2570: Spurious ownership fix on every start when directory group differs from owner's primary group

yorickdowne:
**What I did**

Improve the use case where `node:node-admins` owns the directory, and `alice` and `bob` are members of `node-admins`

`cp` for created files becomes `install`, because I couldn't figure out how to be graceful otherwise when the user has group write permissions but forgot setgid

The base case of owner `alice:alice` and user `alice` has not been impacted. That's quite important, and if review finds that's not so, that needs to be fixed.

There are corner cases where `bob`, if they don't have sudo and `other` can't read on the directory, needs `alice`, who does have `sudo`, to run `ethd` at least once first, so config files are created and given the right permissions.

There is a test script at `tests/test-multiuser.sh` that runs through a number of scenarios and checks ownership and permissions. That includes running `ethd` as `root` and verifying that file ownership does not become `root`.

It's entirely plausible I missed a corner case. Careful review requested.

Also resolves #2570

### DIFF
--- a/.github/workflows/test-ethd.yml
+++ b/.github/workflows/test-ethd.yml
@@ -36,7 +36,7 @@ jobs:
       - name: Set up Docker on macOS
         if: ${{ startsWith(matrix.os, 'macos-') }}
         uses: douglascamata/setup-docker-macos-action@d5ccc6aae0ce23e7700154f5e63cc53e6433ac48
-      - name: Install Expect
+      - name: Prerequisites on Ubuntu
         if: ${{ startsWith(matrix.os, 'ubuntu-') }}
         run: sudo apt-get install -y expect whiptail
       - name: Prerequisites on macOS
--- a/ethd
+++ b/ethd
@@ -8,6 +8,13 @@ __sample_service="consensus"
 __min_env_version=57
 __target_pg=18
 __current_docker=29
+__min_ubuntu=22
+__suggest_ubuntu="24.04 or 22.04."
+__upgrade_ubuntu="24.04: https://gist.github.com/yorickdowne/94f1e5538007f4c9d3da7b22b0dc28a4"
+__min_debian=12
+__suggest_debian="13 or 12."
+__upgrade_debian="13: https://gist.github.com/yorickdowne/3cecc7b424ce241b173510e36754af47"
+__stat_mode_param=()
 __docker_exe="docker"
 __old_docker=0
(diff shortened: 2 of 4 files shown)


yorickdowne:
This'll conflict with #2569 and will need to be rebased after that's merged

yorickdowne:
Cherry-picked "[Respect umask when installing generated files](https://github.com/ethstaker/eth-docker/pull/2602/changes/553cc86b12c1c140f533c76b82c7795d97334fe9)" and "https://github.com/ethstaker/eth-docker/pull/2602/changes/06b52ad6afd813d47c3e210473d3a896f540f56b" from review PR. Rejected "https://github.com/ethstaker/eth-docker/pull/2602/changes/f6076f7704cdf266452800eaf4068ac7c95cd38d"

pietjepuk2 (approved):

PR #2577, opened 2026-04-26
Clean up node exporter configuration

yorickdowne:
**What I did**

Do not collect cpufreq, file too large

Change `/` mount to `/rootfs` to avoid mount overlap with `/host/sys`

Tell node exporter where to find udev data

Better default behavior of text collector by pointing it to a dummy directory in `/tmp`

Rely on `pid: host` for implicit `/proc`. Confirmed working

### DIFF
--- a/ethd
+++ b/ethd
@@ -1798,6 +1798,11 @@ __env_migrate() {
 # shellcheck disable=SC2016
         __value='${CORE_FILES}${CUSTOM_FILES:+:${CUSTOM_FILES}}'
       fi
+      # Remove after Glamsterdam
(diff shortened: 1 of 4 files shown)


ethd
```diff
@@ -1832,6 +1832,11 @@ __env_migrate() {
 # shellcheck disable=SC2016
         __value='${CORE_FILES}${CUSTOM_FILES:+:${CUSTOM_FILES}}'
       fi
+      # Remove after Glamsterdam
```
pietjepuk2, line 1835: I assume for this to be called, we want to bump the ENV version?

yorickdowne: It's not critical, and it'll be picked up whenever .env version increments. That's why I don't increment it here: "Whenever" is good enough

pietjepuk2 (approved):

PR #2584, opened 2026-04-28
Support podman

yorickdowne:
**What I did**

Add logic that detects podman and sets `.env` variables accordingly. A user with both docker and podman installed would be using podman, if the podman socket is available, and Docker otherwise.

Logging in compose files uses `.env` vars, so that Podman has a log tag of `'{{.ImageName}}|{{.Name}}|{{.ID}}'` and the `journald` driver, Ubuntu and Debian only. Defaults remain unchanged. `ethd` sets `LOG_DRIVER` and `LOG_TAG` to Docker or Podman values, unless it detects that there are user-defined values in place.

Add logging options to a few files that should have had them.

### DIFF
--- a/ethd
+++ b/ethd
@@ -143,30 +143,74 @@ __handle_docker() {
   local exitstatus
   local var
   local regex
+  local message
+  local use_podman=0
+  local use_docker=0
 
   if [[ "${__distro}" =~ (debian|ubuntu) ]]; then
     set +e
-    if ! ${__docker_exe} images >/dev/null 2>&1; then
-      if [[ "${__cannot_sudo}" -eq 1 ]]; then
-        echo "Cannot call Docker and cannot use sudo. Please make your user part of the docker group"
-        exit 1
-      fi
-      echo "Will use sudo to access Docker"
-      __docker_sudo="sudo"
+    if dpkg-query -W -f='${Status}' docker-ce 2>/dev/null | grep -q "ok installed" \
+         || dpkg-query -W -f='${Status}' docker.io 2>/dev/null | grep -q "ok installed"; then
+      use_docker=1
+    fi
+    if dpkg-query -W -f='${Status}' podman 2>/dev/null | grep -q "ok installed"; then
+      use_podman=1
     fi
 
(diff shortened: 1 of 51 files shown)


yorickdowne:
Networking in podman is very challenging. This does not currently work well and should be revisited with Podman 6 in late May / early June.

yorickdowne:
Archer tested this, and their findings are:
"I went ahead and compiled the incoming podman & passt/pasta/pesto & netavark/aardvark-dns, and this is what I get now:

`INFO p2p: Connected peers inboundQUIC=166 inboundTCP=11 outboundQUIC=12 outboundTCP=22 target=200 total=211`

This is without modifying anything on eth-docker side, also inter-container communication works as expected

So incoming podman 6 should be good to use for rootless mode, the only change needed would be using the new `containers.conf` option `rootless_network_port_forwarder="pasta"` "

Documentation would need to note that this is v4-only, it does not work with v6 dual-stack

ethd
```diff
@@ -208,23 +262,43 @@ __handle_docker() {
...
     fi
 
     # on macOS, user would set this manually to Colima or Docker Desktop value
-    if __dodocker system info --format '{{.SecurityOptions}}' | grep -qi rootless; then
-      DOCKER_SOCK=/run/user/$(id -u)/docker.sock
-      if [[ "${__distro}" =~ (debian|ubuntu) && "${__docker_exe}" = "docker" && ( "${__docker_major_version}" -lt 29 \
-            || ( "${__docker_major_version}" -eq 29 && "${__docker_minor_version}" -lt 5 ) ) ]]; then
-        __old_rootless_docker=1
+    if [[ "${__docker_exe}" = "podman" ]]; then
+      if [[ "$(__dodocker system info --format '{{.Host.Security.Rootless}}')" = "true" ]]; then
```
pietjepuk2, line 291: Claude found that this deletes the "old docker" #2614 user-facing message / detection. The message code is still there, but the flag is never set. Probably because This podman PR predates 2614.

yorickdowne: Thanks for finding that. I've restored the code.

pietjepuk2 (approved):

PR #2585, opened 2026-04-29
Make sure we have jq before relying on it during config

yorickdowne:
**What I did**

`./ethd config` needs `jq`, but did not check that it was installed

### DIFF
--- a/ethd
+++ b/ethd
@@ -5167,14 +5167,19 @@ __query_mev_factor() {
     return
   fi
 
-  if [[ "${__cannot_sudo}" -eq 0 ]] && ! dpkg-query -W -f='${Status}' speedtest-cli 2>/dev/null | grep -q "ok installed"; then
-    echo "Installing speedtest-cli"
-    ${__auto_sudo} apt-get update && ${__auto_sudo} apt-get install -y --no-install-recommends speedtest-cli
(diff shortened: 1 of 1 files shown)


ethd
```diff
@@ -5146,14 +5146,16 @@ __query_mev_factor() {
...
-    echo "Installing speedtest-cli"
-    ${__auto_sudo} apt-get update && ${__auto_sudo} apt-get install -y --no-install-recommends speedtest-cli
-  fi
-  if ! dpkg-query -W -f='${Status}' speedtest-cli 2>/dev/null | grep -q "ok installed"; then
-    echo "Cannot install \"speedtest-cli\", skipping MEV Build Factor query"
-    return
+  if ! dpkg-query -W -f='${Status}' speedtest-cli 2>/dev/null | grep -q "ok installed" || ! dpkg-query -W -f='${Status}' jq 2>/dev/null | grep -q "ok installed"; then
+    if [[ "${__cannot_sudo}" -eq 0 ]]; then
+      echo "Installing speedtest-cli and jq"
+      ${__auto_sudo} apt-get update && ${__auto_sudo} apt-get install -y --no-install-recommends speedtest-cli jq
```
pietjepuk2: contrived case I guess (pointed out by Codex), but if `update` fails, we continue execution below. If speedtest was already installed for some reason, but jq is not, we might get weird errors.

yorickdowne: Aren't we in `set -e`? We just fail

yorickdowne: Learn something about bash every day. I've wrapped it in if

pietjepuk2 (approved):

PR #2586, opened 2026-04-29
Do not mount $DOCKER_ROOT/containers into Alloy

yorickdowne:
**What I did**

Alloy reads logs directly from the docker socket, there is no need to mount the containers dir in


grafana-cloud.yml
pietjepuk2, line 104: Do you want to keep this line? You're deleting it from the other two

yorickdowne: Here it's for cadvisor, not alloy. I do want to keep this.

yorickdowne: However good point that I missed alloy. Fixed

pietjepuk2 (approved):

PR #2587, opened 2026-05-02
Add a test of ethd version to CI

yorickdowne:
**What I did**

Web3signer distroless was not the first time an `./ethd version` broke. Have CI find that, from now on

### DIFF
--- a/.github/workflows/build-clients.yml
+++ b/.github/workflows/build-clients.yml
@@ -17,7 +17,7 @@ jobs:
       matrix:
         client:
           - env: |-
-              COMPOSE_FILE=deposit-cli.yml
+              CORE_FILES=deposit-cli.yml
               DEPCLI_DOCKERFILE=Dockerfile.source
             test_cl: false
             test_el: false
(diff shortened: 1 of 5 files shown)


yorickdowne:
Needs https://github.com/ethstaker/eth-docker/pull/2588 to be in and this to be rebased on main after, for tests to be able to run

.github/workflows/build-clients.yml
```diff
@@ -29,143 +29,140 @@ jobs:
...
-              COMPOSE_FILE=nimbus-cl-only.yml:nimbus-vc-only.yml:nethermind.yml
+              CORE_FILES=nimbus-cl-only.yml:nimbus-vc-only.yml:nethermind.yml
             test_cl: true
             test_el: true
             test_vc: true
           - env: |-
-              COMPOSE_FILE=prysm.yml:geth.yml
+              CORE_FILES=prysm.yml:geth.yml
               PRYSM_DOCKERFILE=Dockerfile.source
-              CL_NODE=consensus:4000
```
pietjepuk2, line 156: Codex flagged this as "this var is still needed by the VC-only variant".  Not sure if this only shows up on the source builds, something to keep an eye on I guess if they trip.

pietjepuk2 (approved):

.github/workflows/build-clients.yml
yorickdowne: Ah that’s a good point. Yep that is needed by the VC only variants. I’ll fix that

PR #2589, opened 2026-05-03
Add mev-boost source build to CI

yorickdowne:
**What I did**

mev-boost source build to be tested by CI


yorickdowne:
I already had this, hah

PR #2590, opened 2026-05-03
Fix yml default source build target

yorickdowne:
**What I did**

The `eval` for this in `Dockerfile.source` never worked, because the `'` was being passed through literally. Change to `$$`. Tested working.

### DIFF
--- a/besu.yml
+++ b/besu.yml
@@ -14,7 +14,7 @@ services:
       context: ./besu
       dockerfile: ${BESU_DOCKERFILE}
       args:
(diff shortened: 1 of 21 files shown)


pietjepuk2:
`nethermind.yml` still has the same issue I think, but uses `git tag`

```
BUILD_TARGET=${NM_SRC_BUILD_TARGET:-'$(git tag --sort=-committerdate |
```

yorickdowne:
Thanks for catching that! Fixed, and tested working

pietjepuk2 (approved):

PR #2591, opened 2026-05-03
Prysm archive node backfills to slot 0

yorickdowne:
**What I did**

A checkpoint-synced Prysm archive node will backfill to genesis

### DIFF
--- a/prysm/docker-entrypoint.sh
+++ b/prysm/docker-entrypoint.sh
@@ -61,6 +61,9 @@ fi
 # Check whether we should rapid sync
 if [[ -n "${CHECKPOINT_SYNC_URL:+x}" ]]; then
   __checkpoint_sync="--checkpoint-sync-url=${CHECKPOINT_SYNC_URL} --enable-backfill"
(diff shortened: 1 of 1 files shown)


pietjepuk2 (approved):

prysm/docker-entrypoint.sh
```diff
@@ -61,6 +61,9 @@ fi
 # Check whether we should rapid sync
 if [[ -n "${CHECKPOINT_SYNC_URL:+x}" ]]; then
   __checkpoint_sync="--checkpoint-sync-url=${CHECKPOINT_SYNC_URL} --enable-backfill"
+  if [[ "${NODE_TYPE}" = "archive" ]]; then
+    __checkpoint_sync+=" --backfill-oldest-slot 1"
```
pietjepuk2: "Setting a value of 0 allows backfilling to genesis"

https://github.com/OffchainLabs/prysm/pull/13729

So is it 0 or 1?

yorickdowne: It's 0 in testing. Although the help text says it defaults to 0, and the tested default is that it gets the minimum retention

PR #2592, opened 2026-05-04
Fix ethd CI on macOS

yorickdowne:
**What I did**

On macOS, `expect` always failed. The recent stricter test exposed this. Run `expect -d` for debug, which is always useful.


yorickdowne:
No issues found. Maybe it's just the occasional slowdown of the macOS CI? At any rate, having this `-d` will be helpful

PR #2595, opened 2026-05-05
Reth uses the same port for discv4 and discv5

yorickdowne:
**What I did**

Discv5 and discv4 are on the same port. Requires Reth 2.2.0

"Support binding Discv5 and Discv4 to the same port (https://github.com/paradigmxyz/reth/pull/23613)"

Rely on discv5 being enabled by default

"Discv5 is now enabled by default (https://github.com/paradigmxyz/reth/pull/23686)"


yorickdowne:
Tests are failing because Ubuntu update servers are under DDoS

pietjepuk2 (approved):

PR #2599, opened 2026-05-06
Refactor DDNS updater code

yorickdowne:
**What I did**

Refactor the DDNS updater so it can use multiple providers. Add the CloudFlare provider. Add a test harness for the new code, so it can be tested end-to-end

`CF_ZONE_API_TOKEN` is gone, this split is no longer needed: A scoped DNS Edit token now suffices for CloudFlare to be able to get the zone ID from a domain name.

This refactor heavily relied on opencode's "Big Pickle" model, with human review and revision


yorickdowne:
Route 53 test results

--- Test: Route53 basic update cycle (wait 20s) ---
  ✓ AWS identity verified (validate() succeeded)
  ✓ Auth via environment variables
  ✓ A record was updated
  ✓ AAAA record was updated
  ✓ CNAME record was updated

--- Test: Route53 missing credentials (expected failure) ---
  ✓ Gracefully rejected missing credentials

--- Test: Unknown provider (expected failure) ---
  ✓ Gracefully rejected unknown provider

All requested tests complete.

yorickdowne:
CloudFlare test results

--- Test: Cloudflare basic update cycle (wait 20s) ---
  ✓ Cloudflare validation passed
  ✓ A record is already up-to-date (no change needed)
  ✓ AAAA record already up-to-date (no change needed)
  ✓ CNAME record already up-to-date (no change needed)

--- Test: Cloudflare invalid token (expected failure) ---
  ✓ Gracefully rejected invalid token

All requested tests complete.

traefik-utils/app/route53_provider.py
```diff
@@ -0,0 +1,164 @@
...
+            logger.info(f"Using AWS profile '{profile}' from {credentials_path}")
+            logger.debug(
+                f"Using AWS creds file: {os.environ.get('AWS_SHARED_CREDENTIALS_FILE', 'N/A')}"
+            )
+            logger.debug(
+                f"Using AWS config file: {os.environ.get('AWS_CONFIG_FILE', 'N/A')}"
+            )
+            return boto3.session.Session(profile_name=profile.strip())
+        else:
+            raise RuntimeError(
```
pietjepuk2, line 59: What about default profile / credentials? Do you want to require users to be explicit about AWS_PROFILE ?

traefik-utils/app/updater.py
```diff
@@ -227,97 +122,14 @@ def get_external_ip6() -> str | None:
...
-    # Already fully-qualified for this domain
-    if n == d or n.endswith("." + d):
-        return n + "."
-
-    # Some other absolute name (contains a dot), just normalize trailing dot
-    if "." in n:
-        return n + "."
-
-    # Short label: expand with domain
     return f"{n}.{d}."
```
pietjepuk2, line 146: old code did `  if n == d or n.endswith("." + d):`, so if the `cname_label` was already a domain like `something.example.com.`, it would not _again_ append `example.com.`, leading to `something.example.com.example.com.`. Not sure if you want to restore this check.

traefik-utils/app/route53_provider.py
yorickdowne: Yes, I do. I want users to be explicit, and I want the config to be easy to read: This way, it's obvious from `.env` whether it's configured for AWS, and if so how

traefik-utils/app/updater.py
yorickdowne: Yeah fair. I restored one case, which is foo.example.com. I don't want foo.bar.com when I'm in example.com, and I don't want root-level CNAMEs. So if someone does foo.bar.com and we're in example.com then yes that'll become foo.bar.com.example.com - we can't create `foo.bar.com.` in `example.com`, that's just not possible.

Root-level CNAMEs are tricky, they are not real CNAMEs and each provider handles them differently. I don't need the aggro. We're not even doing root-level A/AAAA records right now, though the code could be changed to support them with `@`. I'm not sure it's worth doing tbh, and certainly not in this PR.

pietjepuk2 (approved):

PR #2602, opened 2026-05-09
Pr 2575 Testing

pietjepuk2:


yorickdowne:
So I am on the fence about  "Refuse existing multi-user test accounts". Because the script does stuff like "Stopping here so this can be investigated. Note users have not been deleted yet, so you can investigate the temp directory at $__temp_dir" when there is a test failure.

I've added a `--cleanup` option to handle that

yorickdowne:
"Respect umask" handles a corner case. The user has umask 077 but yet chose to create the directory with 022, not 027. Fair enough. Let me test the impact

yorickdowne:
"Keep multi-user fixtures close to real checkouts" - no absolutely not. This is deliberate. I am testing various scenarios of how a user may have set up their directory after git clone. They might have 644 on files. They might have 640 on files. This is an integral part of the test, and I am not removing it

yorickdowne:
Thank you for careful review!

Issue #2593, opened 2026-05-04
Dockerfile.source will use provided binary or build from source


yorickdowne:
For faster debug iteration. From Grandine team:

One idea I wanted to offer - our devs often needs to debug Grandine instances managed by eth-docker. The problem is that building from source takes ages in eth-docker. We do a hack - we comment out the section that fetches dependencies and builds Grandine, instead we just copy the binary in that Dockerfile (developers scp that binary to the server).

I know it's a bit rare usage, but eth-docker could check is there a binary in special dir (for example "grandine/grandine" or so) and then either copy that binary or build from source.

PR #2611, opened 2026-05-12
Optional provided binary for Grandine source build

yorickdowne:
**What I did**

Grandine can get a `grandine` binary in `grandine/src-built-binary/`, for faster debug iteration. Dockerfile can only copy from the context: An arbitrary location in the file system is not possible

Implements #2593


sauliusgrigaitis:
Looks great, thanks!

sauliusgrigaitis (approved):

PR #2617, opened 2026-05-17
fix: reset slashing protection flags per key in validator-import loop

oksanaphmn:
found_one and do_a_protec were never reset between loop iterations. When importing multiple keys, stale values from a previous key caused the no slashing protection warning to be silently suppressed and
sent slashing_protection: null to the keymanager API.


yorickdowne (approved):

yorickdowne:
Thank you for catching this. That was indeed broken since forever.

PR #2626, opened 2026-06-03
Fix Geth EraE import

yorickdowne:
**What I did**

Geth EraE import did not even work remotely. Revamped to download and then import EraE files; tested on Sepolia


yorickdowne:
- Download works
- sha256sum works, including failing on checksum error
- Import relies on https://github.com/ethereum/go-ethereum/pull/35104 to work

pietjepuk2 (approved):

PR #2629, opened 2026-06-03
Support Nethermind EraE import

yorickdowne:
**What I did**
 
Support new Nethermind `1.38.0` EraE import. Tested successfully on Sepolia with `.erae` files. Note the correct file type is `.ere` and Nethermind may need to adjust when PandaOps transitions, unless Nethermind supports both already.


yorickdowne:
The Geth PR https://github.com/ethstaker/eth-docker/pull/2626 should come in first, otherwise the change from `ERA_URL` to `ERE_URL` will trip it up

pietjepuk2 (approved):

PR #2643, opened 2026-06-17
Support Nimbus EL rolling expiry

yorickdowne:
**What I did**

Rolling expiry requires Nimbus EL `0.3.1` and uses `33_024` epochs, to match the CL retention period

### DIFF
--- a/nimbus-el/docker-entrypoint.sh
+++ b/nimbus-el/docker-entrypoint.sh
@@ -151,6 +151,10 @@ case "${NODE_TYPE}" in
         ;;
     esac
     ;;
+  rolling-expiry)
+    echo "Nimbus EL minimal node with 33,024 epochs rolling expiry - ~5 months"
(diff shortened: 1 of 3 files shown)


nimbus-el/docker-entrypoint.sh
```diff
@@ -151,6 +151,10 @@ case "${NODE_TYPE}" in
         ;;
     esac
     ;;
+  rolling)
```
pietjepuk2: `rolling-expiry`? That's the value set in `default.env`

ethd:3483 still has `if [[ "${target}" =~ ^(pre-prague|rolling|aggressive)$ ]]; then` ; do we need to take 'rolling' out there?

yorickdowne: Yep good catch. Will fix

pietjepuk2 (approved):

PR #2645, opened 2026-06-18
Add pre-provisioned time sync drift alert

yorickdowne:
**What I did**

Alert when time sync drifts more than 0.2s, indicating a sync issue; or max error exceeds 2s, indicating failure.

### DIFF
--- a/grafana/default-alerts/timesync_failed.yml
+++ b/grafana/default-alerts/timesync_failed.yml
@@ -1,0 +1,87 @@
+apiVersion: 1
+groups:
+  - orgId: 1
+    name: Provisioned Alerts
+    folder: Eth Docker
+    interval: 5m
+    rules:
+      - uid: time_sync_failed
+        title: Time Sync Failed
+        condition: C
+        data:
+          - refId: A
+            relativeTimeRange:
+              from: 300
+              to: 0
+            datasourceUid: PBFA97CFB590B2093
+            model:
+              expr: node_timex_maxerror_seconds
+              intervalMs: 15000
+              maxDataPoints: 43200
+              queryType: ""
+              refId: A
+              type: query
+          - refId: B
+            datasourceUid: __expr__
+            model:
+              conditions:
+                - evaluator:
+                    params:
+                      - 0
+                      - 0
+                    type: gt
+                  operator:
+                    type: and
+                  query:
+                    params: []
+                  reducer:
+                    params: []
+                    type: avg
+                  type: query
+              datasource:
+                name: Expression
+                type: __expr__
+                uid: __expr__
+              expression: A
+              intervalMs: 1000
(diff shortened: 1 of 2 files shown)


eth2353:
I think using `node_timex_maxerror_seconds` could be a bit misleading. To my understanding, that metric is the maximum (theoretical?) timesync error, not the real time sync error. This value can grow quickly when time sync servers are not reachable, or can be high in general if a high time sync interval is set. When I check the max values for this metric on some of our servers, I see it reaches 0.4 sometimes and that is under completely normal conditions. So this alert as it is right now could be prone to false positives.

`node_timex_estimated_error_seconds` may be a better fit to see the actual estimated error value.

Also, I would another alert on `node_timex_sync_status != 1` if this is true for more than, say, an hour, indicating time sync is not working. The other metric may trigger by then anyway but still, why not.

yorickdowne:
I'm not using sync_status because of this note:

"Unfortunately some daemons do not handle this flag properly, e.g. chrony-1.30 from Debian/jessie clears STA_UNSYNC flag during daemon initialisation and does not indicate clock synchronization status using this flag. Modern chrony versions should work better. All chrony versions require `rtcsync` option to maintain this flag. OpenNTPD does not touch this flag at all till OpenNTPD-5.9p1."

I'll dig more into estimated_error

yorickdowne:
Thanks for that feedback @eth2353 . I've changed it to two alerts: Clock Drift High when `node_timex_estimated_error_seconds` is >0.2, and time sync failed when `node_timex_maxerror_seconds` > 2.

In testing, when I shut down chrony to simulate a complete daemon failure, `node_timex_estimated_error_seconds` stays flat, and `node_timex_maxerror_seconds` increases linearly.

eth2353:
Lgtm 👍

grafana/default-alerts/timesync_failed.yml
```diff
@@ -0,0 +1,87 @@
...
+        title: Time Sync Failed
+        condition: C
+        data:
+          - refId: A
+            relativeTimeRange:
+              from: 300
+              to: 0
+            datasourceUid: PBFA97CFB590B2093
+            model:
+              expr: node_timex_maxerror_seconds
```
pietjepuk2, line 18: Codex suggested to use `node_timex_sync_status == 0` as an indicator of 'time sync is failing', instead of just looking at seconds deviation.

yorickdowne: Please see discussion as to why I don’t. That requires certain config items to be present, and they might not be

pietjepuk2: Ah yes, OK. MIssed the discussion, just went straight to the code :)

pietjepuk2 (approved):

PR #2646, opened 2026-06-18
Remove Grafana datasource deletion

yorickdowne:
**What I did**

Users have gone through multiple hardforks and updated: No need to delete old Prometheus/Loki data sources any longer

### DIFF
--- a/grafana/datasource.yml
+++ b/grafana/datasource.yml
@@ -1,16 +1,5 @@
-# config file version
 apiVersion: 1
 
-# list of datasources that should be deleted from the database
-deleteDatasources:
-  - name: Prometheus
-    orgId: 1
-
-  - name: Loki
-    orgId: 1
-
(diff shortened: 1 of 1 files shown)


grafana/datasource.yml
```diff
@@ -1,16 +1,5 @@
-# config file version
 apiVersion: 1
 
-# list of datasources that should be deleted from the database
```
pietjepuk2, line 4: This does not affect existing users right, i.e. leaves their copy of this file on the 'grafana-config' docker volume intact. Is that intentional/correct?

yorickdowne: This file is changed for existing users, it's copied in the `Dockerfile`

```
COPY --chown=grafana:root ./datasource.yml /etc/grafana/provisioning/datasources/
```

pietjepuk2 (approved):

PR #2649, opened 2026-06-19
Use sudo for chmod only, not for find

yorickdowne:
**What I did**

If `alice:node-admins` owns the directory and `bob:node-admins` runs `ethd`, then a `sudo` is forced on every run, just for a `find`. Change that to sudo the `chmod` only, and run the `find` as `bob`. This makes running `ethd` commands a whole lot smoother for bob, and works just as well for fixing permissions if needed

image

### DIFF
--- a/ethd
+++ b/ethd
@@ -688,9 +688,9 @@ __prep_conffiles() {
 # Adjust "other" read permissions for files that are bind-mounted
   for dir in "${!chmod_dirs[@]}"; do
 # shellcheck disable=SC2086
-     ${__chmod_sudo} find "${dir}" -type d \! -perm -o+rx -exec chmod o+rx {} +
+     find "${dir}" -type d \! -perm -o+rx -exec ${__chmod_sudo} chmod o+rx {} +
 # shellcheck disable=SC2086
-     ${__chmod_sudo} find "${dir}" -type f -name "${chmod_dirs[${dir}]}" \! -perm -o+r -exec chmod o+r {} +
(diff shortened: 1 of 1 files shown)


pietjepuk2:
Elsewhere we have the pattern

```
# Make sure directory owner owns the SSV dkg output dir and everything in it. ssvlabs/ssv-dkg runs as root
  if ${__as_owner} find ./.eth/dkg_output \( \! -user "${__owner}" -o \! -group "${__owner_group}" \) -print -quit | grep -q .; then
    if [[ "${__cannot_sudo}" -eq 0 ]]; then
      echo "Fixing ownership of .eth/dkg_output"
      ${__auto_sudo} chown -R "${__owner}:${__owner_group}" .eth/dkg_output
      ${__chmod_sudo} chmod u+rwx,o+rx .eth/dkg_output
    else
      echo "Ownership of .eth/dkg_output should be fixed, but this user can't sudo"
    fi
  fi
  ```
  
  Because if `find` runs into a case where it _needs_ sudo, it will fail right? The above block captures that afaict, but not sure about your changes in this PR.

yorickdowne:
For this find I don’t expect the ownership to be potentially off, just the permissions. There is a limit to how much auto fixing ethd will do - things where eth docker itself may cause wrong ownership it’ll fix; things where the user monkeyed with it, let them un-monkey it again

pietjepuk2 (approved):

Issue #2652, opened 2026-06-19
Split CL archive and blob archive


yorickdowne:
New `CL_NODE_TYPE` `blob-archive`, and adjust existing during `./ethd update` - it's unlikely someone meant theirs to be a blob archive, but right now some are.

PR #2654, opened 2026-06-21
Add CL blob-archive type

yorickdowne:
**What I did**

`CL_NODE_TYPE` now has `archive` and `blob-archive` to be more deliberate about whether to keep blobs. Prepares for EraB import.

`./ethd update` will migrate `archive` to `blob-archive` on those clients that support it

**Related issue**
Implements #2652 

image

### DIFF
--- a/ethd
+++ b/ethd
@@ -1980,6 +1980,11 @@ __migrate_env() {
       if [[ "${var}" = "NODE_EXPORTER_COLLECTOR_MOUNT_PATH" && "${__value}" = "/dev/null" ]]; then
         __value=""
       fi
+      # Remove after Glamsterdam
+      if [[ "${__source_ver}" -lt "59" && "${var}" = "CL_NODE_TYPE" && "${__value}" = "archive"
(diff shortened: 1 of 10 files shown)


ethd
```diff
@@ -1980,6 +1980,11 @@ __migrate_env() {
       if [[ "${var}" = "NODE_EXPORTER_COLLECTOR_MOUNT_PATH" && "${__value}" = "/dev/null" ]]; then
         __value=""
       fi
+      # Remove after Glamsterdam
+      if [[ "${__source_ver}" -lt "59" && "${var}" = "CL_NODE_TYPE" && "${__value}" = "archive"
+            && ${COMPOSE_FILE} =~ (prysm\.yml|prsym-cl-only\.yml|lighthouse\.yml|lighthouse-cl-only\.yml|lodestar\.yml|lodestar-cl-only\.yml|caplin\.yml) ]]; then
```
pietjepuk2: prsym-cl-only -> prysm-cl-only

yorickdowne: Thank you. Fixed.

pietjepuk2 (approved):

PR #2657, opened 2026-06-22
Nimbus VP can parse comma-separated URLs

yorickdowne:
**What I did**

Nimbus Verified Proxy handles comma-separated `CL_NODE` correctly, and accepts comma-separated `RPC_URLS`

**Related issue**
closes #2647 

**(not mandatory) A picture of a cute animal, if possible in relation to what you did**

image

PR #2658, opened 2026-06-22
Better handling of comma-separated CL_NODE

yorickdowne:
**What I did**

Better UX for comma-separated `CL_NODE` in more places. `ethdo` now finds a usable `CL_NODE`, as do keymanager and the legacy Lighthouse validator exit. Teku legacy exit uses a comma-separated list, and Siren uses the first entry in the list. I don't have curl in Siren to get fancier.

image

### DIFF
--- a/ethdo/docker-entrypoint.sh
+++ b/ethdo/docker-entrypoint.sh
@@ -83,8 +83,20 @@ if [[ "$*" =~ "--offline" ]]; then
     chown ethdo:ethdo /app/offline-preparation.json
   fi
 else
-  # Get just the first CL_NODE
-  __args=( "${__args[@]:0:1}" "--connection" "$(cut -d, -f1 <<<"${CL_NODE}")" "${__args[@]:1}" )
+  nodes=$(echo "${CL_NODE}" | tr ',' ' ')
+  for node in ${nodes}; do
+    if curl -s -m 5 -o /dev/null -w "%{http_code}" "${node}" | grep -q "^[23]"; then
+      node_reachable=1
+      break
+    fi
+  done
+
+  if [[ "${node_reachable}" -eq 0 ]]; then
+    echo "No consensus client node is reachable via any URL in ${CL_NODE}"
+    exit 1
+  fi
+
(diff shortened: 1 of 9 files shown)


ethdo/docker-entrypoint.sh
```diff
@@ -83,8 +83,20 @@ if [[ "$*" =~ "--offline" ]]; then
     chown ethdo:ethdo /app/offline-preparation.json
   fi
 else
-  # Get just the first CL_NODE
-  __args=( "${__args[@]:0:1}" "--connection" "$(cut -d, -f1 <<<"${CL_NODE}")" "${__args[@]:1}" )
+  nodes=$(echo "${CL_NODE}" | tr ',' ' ')
+  for node in ${nodes}; do
+    if curl -s -m 5 -o /dev/null -w "%{http_code}" "${node}" | grep -q "^[23]"; then
```
pietjepuk2, line 88: Same potential regression as in #2657; only some return codes accepted now (occurs multiple times this PR, not just this one)

siren.yml
```diff
@@ -25,7 +25,12 @@ services:
       - NODE_OPTIONS="--dns-result-order=ipv4first"
     <<: *logging
     entrypoint:
-      - docker-entrypoint.sh
+      - /bin/bash
+      - -c
+    command:
+      - |
+        export BEACON_URL="$${BEACON_URL%%,*}"
```
pietjepuk2: Claude flagged this as "not passing API_TOKEN will break this". 

Also the siren entrypoint already did this, making this change unnecessary:

```
export BEACON_URL=${BEACON_URL%%,*}      # <-- same first-entry strip, since #1978
```

ethdo/docker-entrypoint.sh
yorickdowne: Meant as a fix here as well

siren.yml
yorickdowne: Yep you're right. I completely missed that I had my own docker-entrypoint.sh bind-mounted in. Reverted that change.

pietjepuk2 (approved):

PR #2676, opened 2026-07-02
Support Nimbus Unified client

yorickdowne:
**What I did**

Same code base as Nimbus EL, but configure it to be both CL and EL. VC is standalone, do not use the built-in VC

### DIFF
--- a/ethd
+++ b/ethd
@@ -1403,6 +1403,13 @@ __source_build() {
           __docompose build --pull --no-cache consensus
       fi
       ;;
+    *nimbus-unified.yml*)
+      var="NIMUNIFIED_DOCKERFILE"
+      __get_value_from_env "${var}" "${__env_file}" "__value"
+      if [[ "${__value}" = "Dockerfile-unified.source" ]]; then
+          __docompose build --pull --no-cache consensus
+      fi
+      ;;
     *prysm.yml*|*prysm-cl-only.yml*)
       var="PRYSM_DOCKERFILE"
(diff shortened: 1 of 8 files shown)


yorickdowne:
Does not yet allow separate `--max-peers` for CL and EL, and likely does not work with custom networks. Good enough for testing on Hoodi and Sepolia

yorickdowne:
Tested `ethd resync-consensus` and `ethd resync-execution`

yorickdowne:
There is a Grafana dashboard at https://github.com/status-im/nimbus-eth1/blob/master/metrics/grafana/grafana.infra.status.im.json , but it hard-codes a Prometheus uid. Let's see whether they'll replace with one that doesn't.

https://github.com/status-im/nimbus-eth1/issues/4393

ethd
```diff
@@ -1403,6 +1403,13 @@ __source_build() {
           __docompose build --pull --no-cache consensus
       fi
       ;;
+    *nimbus-unified.yml*)
+      var="NIMUNIFIED_DOCKERFILE"
+      __get_value_from_env "${var}" "${__env_file}" "__value"
+      if [[ "${__value}" = "Dockerfile.source" ]]; then
```
pietjepuk2: Claude caught: `Dockerfile-unified.source`

yorickdowne: Good catch! Resolved.

pietjepuk2 (approved):

PR #2678, opened 2026-07-02
resync-consensus uses uppercase client names

yorickdowne:
**What I did**

Change `ethd resync-consensus` to use upper case client names, in line with `ethd resync-execution`


yorickdowne:
Keep in draft until Nimbus Unified has been merged, then adjust to make Nimbus Unified uppercase as well

pietjepuk2 (approved):

PR #2682, opened 2026-07-02
Use correct syntax for Nimbus EL custom networks

yorickdowne:
**What I did**

Nimbus EL custom path uses `--network`. If `--custom-network` was correct in a previous version, it no longer is

Correct parsing of bootnodes to allow trailing comments, see https://github.com/eth-clients/sepolia/blob/main/metadata/enodes.yaml

### DIFF
--- a/nimbus-el/Dockerfile-unified.binary
+++ b/nimbus-el/Dockerfile-unified.binary
@@ -17,6 +17,8 @@ RUN apt-get update && DEBIAN_FRONTEND=noninteractive TZ=Etc/UTC apt-get install 
   gosu \
   adduser \
   bash \
+  git \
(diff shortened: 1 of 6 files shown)


yorickdowne:
Tested with Sepolia metadata. Works for Nimbus EL. Expected to break for Nimbus Unified: They have an issue to improve things

yorickdowne:
Adjusted to work with https://github.com/status-im/nimbus-eth1/pull/4449

This'll break for Nimbus-Unified as released, but then it was breaking before. This form will start to work when their PR is merged, which is preferable to staying broken

pietjepuk2 (approved):

PR #2684, opened 2026-07-02
Nimbus archive node improvements

yorickdowne:
**What I did**

Do not checkpoint sync when `archive`  and no `ERC_URL`, genesis sync and warn user

`history=archive` also keeps blobs, Nimbus is not granular about state history and blob history

Support Era(C) files with `ERC_URL`

curl and pushd/popd improvements to the downloader, also for ere on Unified. Broader improvement across all entrypoints will be a separate PR


yorickdowne:
Same version as new zkboost, could be merged at the same time

yorickdowne:
Tested with mainnet files at https://mainnet.era.nimbus.team/

pietjepuk2 (approved):

PR #2687, opened 2026-07-03
Add a comment re Lighthouse Web3signer support

yorickdowne:
**What I did**

To avoid future freakouts, add a comment that explains why Lighthouse VC entrypoint has no web3signer parameters. It's been 3 years and I had forgotten that's how it works 😅

### DIFF
--- a/lighthouse/docker-entrypoint-vc.sh
+++ b/lighthouse/docker-entrypoint-vc.sh
@@ -96,6 +96,9 @@ else
   __att_aggr=""
 fi
 
+# Web3signer note: Lighthouse uses the URL given to it by the keymanager remote registration. It does
+# not use a parameter to connect the VC to Web3signer. This is unique to Lighthouse and Vero
+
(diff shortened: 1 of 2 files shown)


lighthouse/docker-entrypoint-vc.sh
```diff
@@ -96,6 +96,9 @@ else
   __att_aggr=""
 fi
 
+# Web3signer note: Lighthouse uses the URL given to it by the keymanager remote registration. It does
+# not use a parameter to connect the VC to Web3signer. This is unique to Lighthouse
+
```
pietjepuk2, line 101: Apparently Vero also works that way? So strictly speaking not the only one?

pietjepuk2 (approved):

lighthouse/docker-entrypoint-vc.sh
yorickdowne: Hmm, I'll take a look, may add a note to Vero as well

PR #2688, opened 2026-07-03
Better VC options in entrypoint, including Graffiti

yorickdowne:
**What I did**

- For "all-in-1" clients, only configure VC options when the embedded VC is in use. This makes it easier for users to understand parameters when they `ps auxww | grep <client-executable>`

- Refactor Graffiti handling. Because spaces in Graffiti have to be preserved, Graffiti was special-cased before, with two separate `exec` lines. Use an array for parameter expansion instead, so `${GRAFFITI}` with spaces remains preserved while also handling spaces between parameters and handling no graffiti parameter when using default graffiti.

### DIFF
--- a/grandine/docker-entrypoint.sh
+++ b/grandine/docker-entrypoint.sh
@@ -188,6 +188,12 @@ else
   __trace=""
 fi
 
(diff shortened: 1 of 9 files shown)


yorickdowne:
Needs to be rebased after https://github.com/ethstaker/eth-docker/pull/2686 and https://github.com/ethstaker/eth-docker/pull/2684 are in

PR #2692, opened 2026-07-05
Prysm source build uses make

### LINKED ISSUE #2670: Switch Prysm source build to `make`, from `bazel`
Once Prysm has the ability to build with `make`, Phase 3/9 of this issue: https://github.com/OffchainLabs/prysm/issues/16982

yorickdowne:
**What I did**

Switch Prysm source build to `make`, from `bazel`

**Related issue**
closes #2670

### DIFF
--- a/prysm/Dockerfile.source
+++ b/prysm/Dockerfile.source
@@ -11,9 +11,6 @@ ARG DOCKER_VC_REPO
 RUN apt-get update && DEBIAN_FRONTEND=noninteractive TZ=Etc/UTC apt-get install -y --no-install-recommends \
   cmake \
   libgmp-dev \
-  npm \
-  && npm install -g @bazel/bazelisk \
-  && bazel version \
   && apt-get clean \
   && rm -rf /var/lib/apt/lists/*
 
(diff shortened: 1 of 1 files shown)


yorickdowne:
Tested successfully on `develop` now that Phase 2/9 is in. For this to work with build CI, Prysm needs a release

pietjepuk2 (approved):

nalepae:
@yorickdowne, I see you now use `make build`, which effectively builds the binaries, but in a non hermetic and reproducible way.

To build hermetic and reproducible binaries, it is advised to use `make dist`, which actually builds everything in a Docker container.

The `make dist` pull request (not yet merged) is:
- https://github.com/OffchainLabs/prysm/pull/17296

yorickdowne:
I’ll take a look at the PR. `make dist` as such won’t work for me because I’m already in a Docker container, but I can look at what it does beyond that.

Also though, none of the client source builds produce reproducible builds. Their intent is not to be used in a release process, but to allow users to try out branches and PRs.

yorickdowne:
`make dist` is described as `Build official release binaries`. That's not what source-building is for. Users who want official release binaries should stick to `Dockerfile.binary`

PR #2705, opened 2026-07-13
Support EIP-8025 zkboost

yorickdowne:
**What I did**
- Support EIP-8025 zkboost optional proofs with experimental Lighthouse build
- Remove previous zk experiment

### DIFF
--- a/default.env
+++ b/default.env
@@ -244,6 +244,8 @@ CONTRIBUTOOR_EXTRAS=
 MEV_EXTRAS=
 # Additional parameters for a verified RPC proxy
 PROXY_EXTRAS=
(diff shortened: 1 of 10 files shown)


yorickdowne:
Open question: Are there mainnet bootnodes that deliver proofs via P2P?

yorickdowne:
"ideally CL would verify the proof locally without `zkboost`" - eventually come Hegotá time this may all be ripped out again. For now it's useful for testing.

yorickdowne:
In draft for now. There are no mainnet P2P nodes to test this with, and the verifier functionality might move into the client itself

PR #2706, opened 2026-07-13
Fix set -e in entrypoint scripts

yorickdowne:
**What I did**

Fix `set -e` in entrypoint scripts. 

Remove `set -e / set +e` from custom network blocks
Add a top-level set to the four scripts that were missing it. 
Three scripts were missing `E`, changed for consistency
Nimbus scripts lose the commented out networkid, it's not needed

### DIFF
--- a/nimbus/docker-entrypoint.sh
+++ b/nimbus/docker-entrypoint.sh
@@ -1,4 +1,5 @@
 #!/usr/bin/env bash
+set -Eeuo pipefail
 
 if [[ "$(id -u)" -eq 0 ]]; then
   chown -R user:user /var/lib/nimbus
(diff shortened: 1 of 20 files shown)


pietjepuk2 (approved):

nimbus/docker-entrypoint.sh
pietjepuk2, line 117: Flagged by Codex:

```
- `MEV_BOOST=true` and `EMBEDDED_VC=false` enters the outer MEV branch.
- `__mev_factor` is assigned only inside the embedded-VC sub-branch.
- Both final `exec` paths expand `${__mev_factor}` under `set -u`.
```

If we just move the initialize up, it should be fine. Or also add this block for an `else` to `"${EMBEDDED_VC}" = "true" `

yorickdowne: Thank you, good catch. Resolved.

pietjepuk2 (approved):

PR #2729, opened 2026-07-24
Besu same port IPv6

yorickdowne:
**What I did**

Besu uses the same port for IPv4 and IPv6, see https://github.com/besu-eth/besu/pull/10624

Load both ENR and enode bootnodes with custom networks

`--discovery-mode BOTH` not enabled yet, it takes high CPU. Default v4 and switch to v5 with IPV6

Full node now uses full sync, because snap sync parameter for full node is deprecated


yorickdowne:
Not in 26.7.1 yet

PR #2745, opened 2026-08-05
Remove pre-commit action and rely on pre-commit.ci

yorickdowne:
**What I did**

Pre-commit action is 2 years old and no longer supported. Switch to pre-commit.ci, which is free for FOSS repos


yorickdowne:
This did not work well, it does not have shellcheck and docker. For now use the old pre-commit action

PR #2751, opened 2026-08-05
Reth source with LLVM 23

yorickdowne:
**What I did**

Bump LLVM to 23

Reth source build needs to be tested before merging


yorickdowne:
This fails. Close, and wait for Reth to require LLVM 23

PR #2816, opened 2026-09-03
ePBS support for Lodestar

yorickdowne:
**What I did**

Support Lodestar's ePBS features in the VC

This requires broad changes across the entire project

High level
- `MEV_BOOST` will run alongside ePBS until Glamsterdam
- Users would need to update Eth Docker again after Glamsterdam to remove MEV Boost

Changes
- Introduce `EPBS_BUILDERS` and `EPBS_BUILDER_URLS`
- `lodestar/docker-entrypoint-vc.sh` handles both `MEV_BOOST` and `EPBS_BUILDERS`
- `MEV_MIN_BID` is renamed `EPBS_MIN_BID` and works for both
- `MEV_BOOST_FACTOR` is renamed `EPBS_BOOST_FACTOR` and works for both
- All clients use the new names

Deliberately did *not* add a variable for trusted bids. Users can seek this with `VC_EXTRAS=--allowDangerousTrustedPayments --builder.maxExecutionPayment <value>`

No support for the new keymanager endpoint in this, yet. That comes in a followup PR


lodestar/docker-entrypoint-vc.sh
```diff
@@ -36,41 +36,55 @@ else
...
       ;;
     *)
-      __mev_factor=""
-      echo "WARNING: MEV_BUILD_FACTOR has an invalid value of \"${build_factor}\""
+      __epbs="--builder"
+      echo "WARNING: EPBS_BUILD_FACTOR has an invalid value of \"${build_factor}\""
       ;;
   esac
+  if [[ -n "${EPBS_MIN_BID}" ]]; then
+    __epbs+=" --builder.minBid ${EPBS_MIN_BID}"
```
nflaig: worth noting that this is in gwei, could also add support for floating point numbers and interpret them as eth values like 0.05, curious if you think that is worth it from a ux perspective since users are used to it already from mev-boost

lodestar/docker-entrypoint-vc.sh
```diff
@@ -36,41 +36,55 @@ else
...
+  build_factor="$(__normalize_int "${EPBS_BUILD_FACTOR}")"
   case "${build_factor}" in
     0)
-      __mev_boost=""
-      __mev_factor=""
-      echo "Disabled MEV Boost because MEV_BUILD_FACTOR is 0."
-      echo "WARNING: This conflicts with MEV_BOOST true. Set factor in a range of 1 to 100"
+      __epbs="--builder.selection executionalways"
+      echo "Disable ePBS builders because EPBS_BUILD_FACTOR is 0."
+      echo "WARNING: This conflicts with EPBS_BUILDERS true. Set factor in a range of 1 to 100"
```
nflaig: worth noting that in case that the local block fails for some reason or is delayed too much we will still fall back to using a bid from p2p

we do have a `--builder.selection executiononly` option which I deprecated for now (it behaves the same as `executionalways`) but I am still debating whether to keep it, how that would work is essentially that it will set a config equivalent to`{"builders": [],"min_bid": "2**64 - 1"}`, at least how lodestar works right now, it disregards any bid below min bid.

the downside with this is that if local el has some issue when producing the payload you might miss the proposal, I am planning to put a dummy bid in the block though, so at least in that case, you still have a valid beacon block, just never reveal the payload

curious if you think that is something worth supporting, I certainly would not wanna advocate for it, but there may be a handful of users that prefer to never take a builder bid? I am not really sure on that

lodestar/docker-entrypoint-vc.sh
yorickdowne: UX wise probably worth it.

thanks for the clarification, I will have our entryscript do the math

lodestar/docker-entrypoint-vc.sh
yorickdowne: Extremely niche, agreed. I’ll adjust the message. I don’t think executiononly is required. Most users would probably prefer some block over no block at all

ethd
```diff
@@ -5518,6 +5528,57 @@ screen.\n\nThe client will not start successfully until a valid ETH rewards addr
...
+    __final_msg+="\nYou are configured to get ePBS block building bids via P2P."
+  fi
+
+  if [[ ! "${NETWORK}" =~ ^(mainnet|hoodi|sepolia|plataberget)$ ]]; then
+    EPBS_BUILDER_URLS=""
+    return 0
+  fi
+
+# No builder URLs known yet
+  EPBS_BUILDER_URLS=""
```
pietjepuk2: Does running config again reset this variable again then, if I set them myself in the env file?

ethd
```diff
@@ -5518,6 +5528,57 @@ screen.\n\nThe client will not start successfully until a valid ETH rewards addr
...
+  if [[ "${MEV_BOOST}" = "false" ]]; then
+    default="--defaultno"
+  else
+    default=""
+  fi
+  if ! whiptail --title "ePBS Builders" --yesno "${message}" 10 65 "${default}"; then
+    EPBS_BUILDERS="false"
+    EPBS_BUILDER_URLS=""
+    if [[ "${MEV_BOOST}" = "false" ]]; then
+      __final_msg+="\nYou are configured to build blocks locally, only.\nPlease ensure you have 200 Mbit/s upload bandwidth or more."
```
pietjepuk2: Elsewhere you have "Build blocks locally, use ePBS builders as fallback"

ethd
```diff
@@ -5518,6 +5528,57 @@ screen.\n\nThe client will not start successfully until a valid ETH rewards addr
...
+    else
+      __final_msg+="\nYou are configured to build blocks locally, only, after Glamsterdam.\nPlease ensure you have 200 Mbit/s upload bandwidth or more."
```
pietjepuk2: Same as above

ethd
yorickdowne: Ah yes it would. Good catch. Let me address that

ethd
yorickdowne: Fallback is a Lodestar behavior. I removed "only", good point. Users need to know they need the bandwidth for local building. Fallback is for failing to even build locally, which would be extremely rare.

ethd
yorickdowne: Logic changed. If the network changed, set it to empty, as we don't know any yet. If it's empty, let the user know they can set it. On networks that won't have well-known relays, deliberately no message. These are likely to be custom networks and users are advanced.

pietjepuk2 (approved):

PR #2819, opened 2026-09-03
Nethermind 2.0 flat is default

yorickdowne:
**What I did**

Prepare for Nethermind 2.0, which offers a FlatDB archive and makes FlatDB the default

I am using this opportunity to deprecate the `nm-eth1-data` volume, which will only ever have been Patricia. I will stop supporting that volume when Nethermind drops Patricia support, version TBD: https://github.com/NethermindEth/nethermind/pull/13409

For a followup-PR with maybe 2.1, consider offering `--FlatDb.ImportFromPruningTrieState=true` mechanism when already on the new volume. 2.1 so the old trie is removed after.


pietjepuk2:
Looked good to me, but Claude had a small finding (suggesting Besu #2824 style fix)

```
Blocker: a new flat archive gets misread as Patricia

I tested this with the 2.0.0-rc2 image on hoodi:
1. I started a fresh archive with the PR's flags and stopped it cleanly at block 0. That is what happens while the consensus client is still syncing.
2. flat/ contained only the write-ahead log and no .sst files, so the PR's check calls the database Patricia.
3. I restarted with the flags the entrypoint would then pass, which leave out HistoryEnabled. Nethermind logged "existing flat DB detected" and kept running as a plain full node. There was no warning, so the archive silently stops recording history.

Fix: use the marker-file approach from the Besu PR (#2824). Record the layout on first start, and use the .sst check only for databases that already exist.
```


suggested patch:
```diff
+# The marker records a DB known to be flat: state persisted before the first RocksDB memtable flush
+# lives in the write-ahead log, so a young flat DB has no SST file to find.
 __patricia=0
-if [[ -d /var/lib/nethermind/nethermind_db || -d /var/lib/nethermind-og/nethermind_db ]]; then
-  if [[ -z "$(find /var/lib/nethermind/nethermind_db /var/lib/nethermind-og/nethermind_db \
-              -mindepth 3 -maxdepth 3 -path '*/flat/*' -name '*.sst' -print -quit 2>/dev/null)" ]]; then
+if [[ ! -f /var/lib/nethermind/flatdb ]]; then
+  if [[ -d /var/lib/nethermind/nethermind_db || -d /var/lib/nethermind-og/nethermind_db ]]; then
+    if [[ -z "$(find /var/lib/nethermind/nethermind_db /var/lib/nethermind-og/nethermind_db \
+                -mindepth 3 -maxdepth 3 -path '*/flat/*' -name '*.sst' -print -quit 2>/dev/null)" ]]; then
+      __patricia=1
+    else
+      touch /var/lib/nethermind/flatdb
+    fi
+  elif [[ "${EL_EXTRAS,,}" =~ (flatdb\.enabled|flatdb-enabled)[[:space:]=]+false ]]; then
     __patricia=1
+  else
+    touch /var/lib/nethermind/flatdb
   fi
-elif [[ "${EL_EXTRAS,,}" =~ (flatdb\.enabled|flatdb-enabled)[[:space:]=]+false ]]; then
-  __patricia=1
 fi
```
Plus one line in ethd's __nethermind_is_patricia, which also unblocks prune-nethermind on a young flat DB:
```diff
   __dodocker run --rm -v "${vol}:/var/lib/nethermind" alpine:3 sh -c '
+    [ -f /var/lib/nethermind/flatdb ] && exit 1
     [ -d /var/lib/nethermind/nethermind_db ] || exit 1
```

yorickdowne:
Concern addressed.

yorickdowne:
Change needed: Check for existence of legacy volume and warn of deprecation. Users could convert to FlatDB and they'd never know something is amiss until I rip out support

yorickdowne:
Nethermind 2.0 expected to release today

PR #2821, opened 2026-09-05
Support Era1 for Besu full node

yorickdowne:
**What I did**

Now that Besu no longer supports `--snapsync-synchronizer-pre-checkpoint-headers-only-enabled=false` for a full node and requires full sync, support `--era1-import-prepipeline-enabled`

This can be replaced by EraE support if Besu enables it. See feature request: https://github.com/besu-eth/besu/issues/11244


besu/docker-entrypoint.sh
```diff
@@ -83,6 +83,12 @@ case "${NODE_TYPE}" in
   full)
     echo "Besu full node without history expiry. Requires \"full\" sync and will take a long time to sync"
     __prune="--sync-mode=FULL --bonsai-limit-trie-logs-enabled=false --snapsync-server-enabled"
+    if [[ -n "${ERE_URL}" ]]; then
```
pietjepuk2: Should we have a `! "${NETWORK}" =~ ^https?://` guard like the other docker entrypoint files?

yorickdowne: For consistency: yes

pietjepuk2 (approved):

PR #2830, opened 2026-09-15
Fix geth attach and volume regexes

yorickdowne:
**What I did**

I am testing for volume existence with `[[ -n "$(__dodocker volume ls -q -f "name=${vol}")" ]]` and analogous for non-existance `-z`. This is not ideal because it's a regex and can give me false positives. `docker inspect` is better for just existence. For volume ID a bit more is needed

Examples: Directory `eth.docker` doesn't match any volumes it should but will match `eth-docker`; directories `eth-docker` and `test-eth-docker` will overmatch volume names inside `eth-docker`

Geth attach hard-codes an old volume name and always failed


pietjepuk2 (approved):

pietjepuk2:
Claude found some minor non-blocking issues:
```
1. Medium, reproduced: the tests/test-traefik.sh cleanup no longer removes anything. compose run names its containers <project>-ddns-run-<hash>, so the new name=^ddns-run filter matches none of them. Timed-out containers now leak. Revert the anchor or use -ddns-run-.
2. Low, reproduced: some project names come back wrapped in quotes. A directory named 123 gives name: "123", and the quotes end up in __compose_project. The quoted name is then cached as exact. Every volume check says the volume is absent, so resync and terminate report nothing to do. Stripping the quotes in the awk fixes it. __set_alloy_project_name has the same flaw.
```

don't think we have numeric-only project names.

yorickdowne:
Ah Claude is being thorough, awesome. Both findings addressed, and also added macOS portability code for awk/gawk

Numeric-only project names are unusual, and completely in the hands of the user. Someone can clone eth docker into `1337` and run it there.

PR #2832, opened 2026-09-17
Support Grandine VC

yorickdowne:
**What I did**

Support new Grandine VC

Not implemented in Grandine VC as of 9/22:
- ePBS
- Support for Obol / distributed

Branch at https://github.com/grandinetech/grandine/tree/feature/separate-vc


yorickdowne:
In draft until Grandine has a release

grandine-vc-only.yml
```diff
@@ -0,0 +1,118 @@
...
+      - --beacon-node-urls
+      - ${CL_NODE}
+      - --metrics
+      - --metrics-address
+      - 0.0.0.0
+      - --metrics-port
+      - "8009"
+      - --suggested-fee-recipient
+      - ${FEE_RECIPIENT}
+      - --track-liveness
```
pietjepuk2, line 65: apparently this is a BN option, not a VC option

grandine-allin1.yml
```diff
@@ -33,6 +33,8 @@ services:
       - CHECKPOINT_SYNC_URL=${CHECKPOINT_SYNC_URL}
       - JWT_SECRET=${JWT_SECRET}
       - MEV_BOOST=${MEV_BOOST}
+      - EPBS_BUILDERS=${EPBS_BUILDERS}
```
pietjepuk2: this yml still uses `docker-entrypoint.sh` , but that one does not do anything with these options?

grandine/docker-entrypoint-vc.sh
```diff
@@ -0,0 +1,123 @@
...
+__normalize_int() {
+  local v=$1
+  if [[ "${v}" =~ ^[0-9]+$ ]]; then
+    v=$((10#${v}))
+  fi
+  printf '%s' "${v}"
+}
+
+
+if [[ ! -f /var/lib/grandine-vc/wallet-password.txt ]]; then
```
pietjepuk2, line 19: `ethd keys get-grandine-wallet` still reads (only) from `grandine`, not `grandine-vc`.

grandine-allin1.yml
yorickdowne: Yes that's a bit by design right now. I don't know what to do with it. So I stuck them there so I have them later when I know what the shape of ePBS is in Grandine. They might not have it yet.

See https://github.com/grandinetech/grandine/issues/9#issuecomment-5708593492

grandine/docker-entrypoint-vc.sh
yorickdowne: So it does. Fixed. Thanks!

grandine-vc-only.yml
yorickdowne: It's both. The VC sends the fee recipient it wants; the BN sets a fallback recipient in case the VC doesn't

grandine-allin1.yml
yorickdowne: I'm assuming they don't have ePBS and removed the supporting code

yorickdowne: Latest is that ePBS is work in progress. If it lands by the time their VC ships, I can revert the commit that rips it out, and adjust parameters to what they'll be then.

PR #2833, opened 2026-09-17
Increase Besu heap for Era import

yorickdowne:
**What I did**

Besu heap needs to be more than 5g for Era1 import.

In draft until right amount determined by testing


yorickdowne:
It's been stable for days with `-Xmx6g`. Merge

PR #2839, opened 2026-09-18
Add ethd port-check command

yorickdowne:
**What I did**

Add an `ethd port-check` command to help users see whether they have inbound peers on the CL, whether they are behind CGNAT, and whether IPv6 dual-stack works.

This gets more important with QUIC default and Mplex TCP deprecated in Q1 2027

I am less exercised about the EL, which is why this only tests the CL.

I can be convinced to add a separate EL command, maybe with a flag. It's pretty busy output as-is and I don't want to do both at the same time.

Code written by Claude Opus 5


yorickdowne:
Example output when all is well

```
Nimbus port check

  IPv4         <public-v4>, discv5 UDP 9000, QUIC UDP 9001, TCP 9000
  IPv6         no address advertised yet
  CGNAT        no signs of it
  Dual-stack   up, but not advertised in discv5
  Peer ID      16Uiu2HAkuZ3KJDszKH3rKns4hrRHLeYB2MzwsvnZqrSYhmb4W9Zq
  Public key   01e330cfe26764ec5ed416e6e950a8b4c3b7d7977a099b639b504dc56969d188
               d2f03fa717c03f6faae60927696b97d02825ee309ef276e0428cf830f95f8d1e
  Inbound      working - 143 peers dialed you, you dialed 17
  Peers        direction   QUIC v4  QUIC v6   TCP v4   TCP v6
               inbound           3       53       87        0
               outbound          0        0       16        1
               counts as your client reports them
```

yorickdowne:
Example output when there are no incoming peers, so user can troubleshoot their port forwarding

```
Teku port check

  IPv4         <public-v4> (behind NAT), discv5 UDP 59090, QUIC UDP 9021, TCP 9020
  IPv6         <public-v6>, discv5 UDP 45785, QUIC UDP 9011
  CGNAT        no signs of it
  Dual-stack   running
  Peer ID      16Uiu2HAmRuESq6aLBWweRA3Wt69BouEnJ9JAW2nVE9Vmw5e36Tye
  Public key   c4cdffcd3bd76e72b0c27667fc85b1f7503d51db2e76e2797ccd19f17a48bb85
               de82dc4f9e0aba8d31d2ab59878813d3b589274dd3b30a647f30559ec288368b
  Inbound      not working - no peer has dialed you, you dialed 0
  Peers        direction   QUIC v4  QUIC v6   TCP v4   TCP v6
               inbound           0        0        0        0
               outbound          0        0        0        0
               counts as your client reports them

Teku is not peering at all. It may still be starting up, or outbound
UDP 9020 is blocked as well.

To fix inbound connectivity:
  - Forward UDP 9020 (discv5) and UDP 9021 (QUIC) to this host.
    The consensus layer needs UDP. A TCP-only forward will not do.
  - If your ISP has you behind carrier-grade NAT, no forward can work. Enabling
    IPv6 with "ethd config" is then the way to accept inbound connections.
  - A node that started in the last few minutes has not had time to be dialed.

No web site can test a UDP port for you - the ones that offer to only speak TCP.
A green result from one of those proves nothing about UDP 9020 or 9021.

To probe discv5 from outside - a VPS, a phone hotspot or a friend's machine
with Docker - and leave nothing behind there:

Run these from a machine with a public IP address, not from this one and not
from anything else on your LAN. A probe sourced from a private address -
10.x, 172.16-31.x, 192.168.x, or a Docker bridge - reaches your node with a
source the Internet could never route a reply to, and some consensus clients
drop it without a word. It will read as a closed port when the port is fine.

Teku advertises discv5 UDP 59090, the port its NAT mapping happens
to use. That is not the port you forward, and it can change. The port to
forward - and the one this pings - is 9020, CL_P2P_PORT in ".env".

Teku's record advertises <public-v4>, which is the address this pings:

  docker run --rm -e IP="<public-v4>" -e PORT="9020" -e PUBKEY="c4cdffcd3bd76e72b0c27667fc85b1f7503d51db2e76e2797ccd19f17a48bb85de82dc4f9e0aba8d31d2ab59878813d3b589274dd3b30a647f30559ec288368b" ethereum/client-go:alltools-latest sh -c '
    out=$(devp2p discv5 ping "enode://${PUBKEY}@${IP}:${PORT}" 2>&1 | tail -n1)
    if [ "$out" = "<nil>" ]; then echo "PASS - $IP:$PORT is reachable over discv5"; else echo "FAIL - $IP:$PORT - $out"; fi
  '

To probe QUIC the same way:

  docker run --rm -e IP="<public-v4>" -e PORT="9021" -e PIP_ROOT_USER_ACTION=ignore -e DEBIAN_FRONTEND=noninteractive python:3-slim sh -c '
    apt-get update -qq &&
    apt-get install -y -qq curl >/dev/null &&
    pip install --quiet --no-cache-dir aioquic cryptography tqdm &&
    curl -sL https://raw.githubusercontent.com/bojanisc/quicmap/main/quicmap.py -o /tmp/q.py &&
    out=$(python3 /tmp/q.py -p "$PORT" "$IP" 2>&1)
    echo "$out"
    if echo "$out" | grep -q endpoint; then echo; echo "PASS - $IP:$PORT responded to QUIC"; else echo; echo "FAIL - $IP:$PORT did not respond to QUIC"; fi
  '

The raw quicmap output prints above; PASS/FAIL summarizes it. An "ALPN: ?"
line is expected - consensus clients speak the "libp2p" ALPN, which quicmap
does not probe for; "server_versions" appearing at all is what proves the
forward works.

Your record also advertises IPv6 <public-v6>. Run these from a probing host with a
global IPv6 address of its own - the same rule as above, a ULA or link-local
source will not do. They create a small IPv6-enabled docker network for the
probe, since Docker's default bridge has no IPv6 route out:

  docker network create --ipv6 ethd-v6-probe >/dev/null 2>&1; docker run --rm --network ethd-v6-probe -e IP="[<public-v6>]" -e PORT="45785" -e PUBKEY="c4cdffcd3bd76e72b0c27667fc85b1f7503d51db2e76e2797ccd19f17a48bb85de82dc4f9e0aba8d31d2ab59878813d3b589274dd3b30a647f30559ec288368b" ethereum/client-go:alltools-latest sh -c '
    out=$(devp2p discv5 ping "enode://${PUBKEY}@${IP}:${PORT}" 2>&1 | tail -n1)
    if [ "$out" = "<nil>" ]; then echo "PASS - $IP:$PORT is reachable over discv5"; else echo "FAIL - $IP:$PORT - $out"; fi
  '

  docker network create --ipv6 ethd-v6-probe >/dev/null 2>&1; docker run --rm --network ethd-v6-probe -e IP="<public-v6>" -e PORT="9011" -e PIP_ROOT_USER_ACTION=ignore -e DEBIAN_FRONTEND=noninteractive python:3-slim sh -c '
    apt-get update -qq &&
    apt-get install -y -qq curl >/dev/null &&
    pip install --quiet --no-cache-dir aioquic cryptography tqdm &&
    curl -sL https://raw.githubusercontent.com/bojanisc/quicmap/main/quicmap.py -o /tmp/q.py &&
    out=$(python3 /tmp/q.py -p "$PORT" "$IP" 2>&1)
    echo "$out"
    if echo "$out" | grep -q endpoint; then echo; echo "PASS - [$IP]:$PORT responded to QUIC"; else echo; echo "FAIL - [$IP]:$PORT did not respond to QUIC"; fi
  '

None of these commands carry your ENR, so blanking out the IP value is enough to
share them when you ask for help. Your ENR does contain your address - it is not
safe to paste in its place.
```

pietjepuk2 (approved):

PR #2849, opened 2026-09-23
Add ethd keys CI

yorickdowne:
**What I did**

Store 3 Hoodi pubkeys
Add CI for `ethd keys`, the `keymanager.sh` code, which had only manual testing up to now

Features added to production code:
- Prysm and Geth support named `ephemery` network
- `validator-keys` no longer `depends_on`. This never worked as intended to bring the validator service up in time, as it needs a few minutes to have the keymanager API. This cuts down the time to start the validator-keys container.

Workarounds in production code:
- Allow `200` for Prysm graffiti POST and DELETE
- Allow `204` for Teku graffiti POST
- Handle `200` for unknown builder path in Prysm

Bugs fixed in production code:
- keymanager can decode hex-encoded graffiti
- ethd doesn't output warnings when using `--json`
- sign-exit skips keys that don't have an index
- Docker space display on macOS needs `gawk`
- `keys prepare-address-change` now uses `curl` so it works on macOS, and checks whether `curl` exists on Debian and Ubuntu

Bugs filed upstream:
- Grandine returns error on duplicate
- Prysm returns `200` for Graffiti POST and DELETE
- Prysm returns `200` for unknown paths
- Teku returns `204` for Graffiti POST
- Besu cannot sync Ephemery
- Lodestar VC cannot connect to CL on Ephemery, if started a little later

Question asked upstream:
- Should spec be clarified for graffiti GET? Lighthouse returns hex-encoded ASCII

Code by Opus 5.5


yorickdowne:
- [ ] macOS
- [ ] test-grandine
- [ ] test-grandine-plugin
- [ ] test-lighthouse
- [x] test-lodestar
- [ ] test-nimbus
- [ ] test-prysm
- [ ] test-teku

